PR #14, opened 2021-04-28
Create FlatLink

hungqt:
Creating a FlatLink component

Screenshots:
image
image
image

### DIFF
--- a/explorer/src/Explorer.elm
+++ b/explorer/src/Explorer.elm
@@ -6,6 +6,7 @@ import Stories.Dropdown as Dropdown
 import Stories.NumberInput as NumberInput
 import Stories.RadioButton as RadioButton
 import Stories.TextInput as TextInput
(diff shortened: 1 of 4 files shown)


fshauge:
Should the [disabled](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/disabled) attribute be set as well?

hungqt:
> Should the [disabled](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/disabled) attribute be set as well?

FlatLinks are [a-elements](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a), and they don't have a `disabled` attribute unfortunately. That's why i had to create a `withDisabled` function to put disabled into the config if needed.

fshauge:
Aha! Didn't know that. Looks good then!

fshauge (approved):

PR #20, opened 2021-05-28
Add accordion component

fshauge:
image


fshauge:
Note: Accordions are also possible to implement using `<input type="checkbox">` without using any form of Elm model/view/update. The most optimal approach would be to use `<details>` + `<summary>`, but these elements are not supported in IE11.

hungqt (approved):
Looks good 🙌

PR #22, opened 2021-06-16
Support for FlatLink styled as a Button

hungqt:
By using the `withButtonStyles` function for FlatLink, we can now have a-tags styled as buttons

image


fredrikloberg (approved):
💯

PR #24, opened 2021-06-28
Add error border line on text input

mathilwa:
As a step closer to getting full error handling on text input I have added red border when withError is set to True.

Screen Shot 2021-06-28 at 8 59 05 AM
Screen Shot 2021-06-28 at 8 59 10 AM

### DIFF
--- a/src/Nordea/Components/TextInput.elm
+++ b/src/Nordea/Components/TextInput.elm
@@ -23,7 +23,7 @@ type alias Config msg =
     { value : String
     , onInput : Maybe (String -> msg)
     , placeholder : Maybe String
-    , error : Maybe Bool
+    , showError : Bool
     }
 
 
@@ -37,7 +37,7 @@ init value =
         { value = value
         , onInput = Nothing
         , placeholder = Nothing
-        , error = Nothing
+        , showError = False
         }
 
 
@@ -52,7 +52,7 @@ withPlaceholder placeholder (TextInput config) =
 
 withError : Bool -> TextInput msg -> TextInput msg
 withError condition (TextInput config) =
-    TextInput { config | error = Just condition }
+    TextInput { config | showError = condition }
 
 
 
@@ -79,17 +79,9 @@ getAttributes config =
 getStyles : Config msg -> List Style
 getStyles config =
     let
-        borderColorNormal =
-            Colors.grayMedium
         borderColorStyle =
-            case config.error of
-                Just error ->
-                    if error then Colors.redDark
-                    else borderColorNormal
-
-                Nothing ->
-                    borderColorNormal
-
+            if config.showError then Colors.redDark
+            else Colors.grayMedium
     in
     [ fontSize (rem 1)
         , height (em 2.5)
--- a/src/Nordea/Components/TextInput.elm
+++ b/src/Nordea/Components/TextInput.elm
@@ -7,7 +7,7 @@ module Nordea.Components.TextInput exposing
     , withError
     )
 
-import Css exposing (Style, backgroundColor, border2, borderBox, borderColor, borderRadius, boxSizing, disabled, em, focus, fontSize, height, none, outline, padding2, pct, rem, solid, width)
+import Css exposing (Style, backgroundColor, border3, borderBox, borderColor, borderRadius, boxSizing, disabled, em, focus, fontSize, height, none, outline, padding2, pct, rem, solid, width)
 import Html.Styled exposing (Attribute, Html, input, styled)
 import Html.Styled.Attributes exposing (placeholder, value)
 import Html.Styled.Events exposing (onInput)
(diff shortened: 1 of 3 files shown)


src/Nordea/Components/TextInput.elm
```diff
@@ -43,6 +23,7 @@ type alias Config msg =
     { value : String
     , onInput : Maybe (String -> msg)
     , placeholder : Maybe String
+    , error : Maybe Bool
```
fredrikloberg, line 26: Is the meaning clearer without a maybe? 🤔 In this case we don't have to question the difference between `Nothing` and `Just False`

```suggestion
    , showError : Bool
```

fredrikloberg (approved):
🙌

src/Nordea/Components/TextInput.elm
```diff
@@ -89,25 +75,36 @@ getAttributes config =
...
+            borderColor Colors.grayMedium
+        borderColorStyle =
+            case config.error of
+                Just error ->
+                    if error then borderColor Colors.redDark
+                    else borderColorNormal
+
+                Nothing ->
+                    borderColorNormal
+
```
fshauge, line 92:
```suggestion
    let
        borderColor =
            if config.error then
                Colors.redDark

            else
                Colors.grayMedium
```

src/Nordea/Components/TextInput.elm
```diff
@@ -89,25 +75,36 @@ getAttributes config =
...
-        [ backgroundColor Colors.grayWarm
-        ]
-    , focus
-        [ outline none
-        , borderColor Colors.blueNordea
+        , height (em 2.5)
+        , padding2 (em 0.75) (em 0.75)
+        , borderRadius (em 0.125)
+        , border2 (em 0.0625) solid
+        , borderColorStyle
```
fshauge, line 99:
```suggestion
        , border3 (em 0.0625) solid borderColor
```

src/Nordea/Components/TextInput.elm
mathilwa: Would this mean that every text input implementet now need to have withError False to it? 
Or will this mean that error/showError: Bool and that it is always default set to False in the model instead of Nothing?

src/Nordea/Components/TextInput.elm
mathilwa: this gave me mismatch error since config.error is Maybe Bool, not Bool

src/Nordea/Components/TextInput.elm
fredrikloberg: I think it makes sense to set it to False by default, and then use `withError True` in cases where it should be true

fshauge: I agree ^

src/Nordea/Components/TextInput.elm
mathilwa: I had to keep the variable name, since borderColor is an imported attribute from CSS

src/Nordea/Components/TextInput.elm
mathilwa: Agree! Changed it now 😊

src/Nordea/Components/TextInput.elm
mathilwa: It fixed itself when removing Maybe from error 😊

fshauge: Yes should have mentioned that 😅

src/Nordea/Components/TextInput.elm
mathilwa: @fredrikloberg @fshauge should withError be renamed to showError as well, or should we keep it as is?

src/Nordea/Components/TextInput.elm
fshauge: Yet another reason for not exposing imported variables? (looking at you Css)

mathilwa: Haha, I agree. It's also easier to get a quick overview of whats imported and whats our own variables when using Css.blabla instead of just blabla. I think 😊

src/Nordea/Components/TextInput.elm
fredrikloberg: I think both are fine

src/Nordea/Components/TextInput.elm
```diff
@@ -89,25 +75,27 @@ getAttributes config =
...
+        , border3 (em 0.0625) solid borderColorStyle
+        , boxSizing borderBox
+        , width (pct 100)
+        , disabled
+            [ backgroundColor Colors.grayWarm
+            ]
+        , focus
+            [ outline none
+            , borderColor Colors.blueNordea
+            ]
```
fshauge, line 99: Seems like the formatting is a bit off. Could you run elm-format again? (note: you have to enable elm-format for every project in IntelliJ IDEA 😉 )

mathilwa: 👀 for details! I'll fix!

fshauge:
Nicely done! 🙌

PR #38, opened 2021-09-22
Add spinner

hungqt:
Added the spinner used in Partnerhub.

Usage:
```
Spinner.default[attributes]
Spinner.small[attributes]
```

https://user-images.githubusercontent.com/14363025/134309858-00f6bd74-23e9-437a-bd70-5637c9100292.mov

image

### DIFF
--- a/src/Nordea/Components/Spinner.elm
+++ b/src/Nordea/Components/Spinner.elm
@@ -1,11 +1,53 @@
-module Nordea.Components.Spinner exposing (..)
+module Nordea.Components.Spinner exposing (default, large, medium, small)
 
+import Css exposing (Style, color, height, rem, width)
 import Html.Styled exposing (Attribute, Html, div)
+import Html.Styled.Attributes exposing (css)
+import Nordea.Resources.Colors as Colors
 import Nordea.Resources.Icons as Icons
 
 
-view : List (Attribute msg) -> Html msg
-view attrs =
+default : List (Attribute msg) -> Html msg
+default attrs =
     div
-        attrs
+        (css [ color Colors.blueDeep ] :: attrs)
+        [ Icons.spinner ]
+
+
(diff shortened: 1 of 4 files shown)


src/Nordea/Components/Spinner.elm
```diff
@@ -0,0 +1,11 @@
+module Nordea.Components.Spinner exposing (..)
+
+import Html.Styled exposing (Attribute, Html, div)
+import Nordea.Resources.Icons as Icons
+
+
+view : List (Attribute msg) -> Html msg
+view attrs =
+    div
+        attrs
```
fredrikloberg, line 10: Should we set Colors.blueDeep as the default color and a default size?

fredrikloberg (approved):
🙌

src/Nordea/Components/Spinner.elm
fshauge: The more (sane) defaults, the better imo! We might even specify sizes in s, m, l, etc. since it's probably not going to vary _that_ much.

hungqt: I agree blueDeep should be the default color yeah. I can create a default with the original size and a small with the 3.75rem size.

hungqt:
Added S/M/L sizes now and defaulted the color to blueDeep. Idk if it made it better or not? @fredrikloberg @fshauge

fredrikloberg:
In PartnerHub, we have only used the same size all over. So what do you think about only adding one size for now and then we can add other ones later if needed? May reduce clutter and bloating of the library 🤷‍♂️

fshauge:
image

Ehhhh... 😅

hungqt:
Yeah the original size doesnt have any size constraint to it. Therefore we use the small one 😂

fshauge:
Looks good!

PR #40, opened 2021-10-04
Remove unnesessary padding

mathilwa:
Remove padding coming naturally from list, which will center the step indicator better

Screen Shot 2021-10-04 at 10 58 01 AM

### DIFF
--- a/src/Nordea/Components/StepIndicator.elm
+++ b/src/Nordea/Components/StepIndicator.elm
@@ -4,7 +4,41 @@ module Nordea.Components.StepIndicator exposing
     , view
     )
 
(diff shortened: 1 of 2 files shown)


src/Nordea/Components/StepIndicator.elm
```diff
@@ -121,6 +87,7 @@ listStyles =
     [ displayFlex
     , listStyleType none
     , alignItems center
+    , paddingLeft (px 0)
```
fredrikloberg, line 90: In PartnerHub, we try to never use px. Doesn't really matter for 0 though 🙂
```suggestion
    , paddingLeft (rem 0)
```

fredrikloberg (approved):
🙌

src/Nordea/Components/StepIndicator.elm
mathilwa: ofc! I fix 🙌

PR #53, opened 2021-11-12
WIP: Temporary changes to components to enable usage in FF and create git hash

Eivinddh:
### **!! This is not meant as an actual change to the component library !!**

This PR is only ment to temporarily change the components to enable usage in FinansFront, and these changes will seperately be addressed in spesific PRs ☺️ It will in addition allow us to use this hash for now, but it shouldn't live for long 🤞


Eivinddh:
This is not needed anymore 🙌

PR #54, opened 2021-11-15
PAR-673: Add card button.

vetlesolgaard:
- Adds a "card" button.
- This is very much a WIP, so I was hoping we could merge it, and extend it later. Design does not seem to have landed yet on how it will look in the end. Only that we'll have a card button component.

Example of how it will be usen in PartnerHub 👇
Screenshot 2021-11-15 at 08 10 18


hungqt (approved):
Nice 💯

fredrikloberg (approved):
Nice 💯

PR #56, opened 2021-11-16
Status component

Eivinddh:
Status component with custom text 🎉 
bilde

### DIFF
--- a/src/Nordea/Components/Status.elm
+++ b/src/Nordea/Components/Status.elm
@@ -1,0 +1,77 @@
+module Nordea.Components.Status exposing (blue, gray, green, red, yellow)
+
+import Css exposing (Color, backgroundColor, borderRadius, display, ellipsis, fontSize, hidden, inlineBlock, lineHeight, maxWidth, overflow, padding, pct, rem, textOverflow)
+import Html.Styled as Html exposing (Html)
+import Html.Styled.Attributes exposing (css)
+import Nordea.Resources.Colors as Color
+
+
+type StatusColor
+    = Green
+    | Red
+    | CloudBlue
+    | Yellow
+    | LightGray
+
+
+view : String -> StatusColor -> Html msg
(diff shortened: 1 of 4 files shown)


src/Nordea/Components/Status.elm
```diff
@@ -0,0 +1,77 @@
...
+type StatusColor
+    = Green
+    | Red
+    | CloudBlue
+    | Yellow
+    | LightGray
+
+
+view : String -> StatusColor -> Html msg
+view text statusColor =
```
fredrikloberg, line 18: I'm a fan of this type of interface. It allows setting arbitrary attributes (e.g., adding a custom css margin, etc.), and allows the content to be set to custom html. Setting custom html instead of a String is handy in cases where we want to set some custom styling to the text, for example. 

```suggestion
view : List (Attribute msg) -> List (Html msg) -> StatusColor -> Html msg
view attrs children statusColor =
```

fredrikloberg (approved):
Looks very good 🤩

src/Nordea/Resources/Colors.elm
```diff
@@ -11,17 +12,21 @@ module Nordea.Resources.Colors exposing
...
     , grayLight
+    , grayLight70
     , grayMedium
     , grayNordea
     , grayWarm
     , green
+    , green70
     , greenDark
     , red
+    , red70
```
mathilwa, line 23: Could these new colors get more descriptive names like the other colors?

src/Nordea/Components/Status.elm
Eivinddh: According to Espen and Thea it should not be possible to change the styling and content ☺️

src/Nordea/Resources/Colors.elm
Eivinddh: Great idea! 🎉

PR #58, opened 2021-11-16
Add card with title and new color to component library

mathilwa:
Add a card with title, separation line and content to component library. To be used with or without shadow. 

Needed to add a new color to the component library, grayCool. 

We tried not to be to strict setting width/maxwidth, because we though it that might be better to set when you are using this card. 

Screen Shot 2021-11-16 at 11 45 59 AM
Screen Shot 2021-11-16 at 11 45 52 AM

### DIFF
--- a/explorer/src/Explorer.elm
+++ b/explorer/src/Explorer.elm
@@ -6,6 +6,7 @@ import Html.Styled exposing (toUnstyled)
 import Nordea.Resources.Fonts as Fonts
 import Stories.Accordion as Accordion
 import Stories.Button as Button
+import Stories.Card as Card
 import Stories.Checkbox as Checkbox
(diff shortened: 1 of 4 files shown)


Eivinddh (approved):
Awesome 🤩 
Well Done LGTM

hungqt (approved):
💯
image

fredrikloberg (commented):
In PartnerHub, we have some cards without a header and separation line. To support that, we have the following interface:

```elm
Card.view
        []
        [ Card.header [] [ text "Card header" ]
        , Card.separationLine
        , ...
        ]
```

Should this card component support scenarios without headers?

fredrikloberg (approved):

PR #59, opened 2021-11-16
Added radio groups and moved some stuff from Label

hhaga:
Added new component `RadioGroup` to use for grouped radio buttons.
To avoid wrapping of all groups in `Label`s we moved `bottomInfo` and `topInfo` out of `Label` and into a new `Common.elm` .
The thought is to use `topInfo` and `bottomInfo` inside the individual components instead of wrapping the components in `Label` where they are used.

To clarify: `topInfo` and `bottomInfo` will still only be used inside the `elm-components library`

### DIFF
--- a/src/Nordea/Components/Common.elm
+++ b/src/Nordea/Components/Common.elm
@@ -1,6 +1,27 @@
 module Nordea.Components.Common exposing (..)
 
-import Css exposing (alignItems, auto, center, color, displayFlex, flex, flexBasis, fontFamilies, fontSize, justifyContent, lineHeight, marginBottom, marginLeft, marginRight, marginTop, none, pct, rem, spaceBetween)
+import Css
+    exposing
+        ( alignItems
+        , auto
+        , center
+        , color
+        , displayFlex
+        , flex
+        , flexBasis
+        , fontFamilies
+        , fontSize
+        , justifyContent
+        , lineHeight
+        , marginBottom
+        , marginLeft
+        , marginRight
+        , marginTop
+        , none
+        , pct
+        , rem
+        , spaceBetween
+        )
 import Html.Styled as Html exposing (Attribute, Html, div, styled, text)
 import Html.Styled.Attributes exposing (css)
 import Maybe.Extra as Maybe
@@ -31,6 +52,8 @@ type alias CharCounter =
     }
 
 
+{-| Should be replaced by common typography
+-}
 bodyText : List (Attribute msg) -> List (Html msg) -> Html msg
 bodyText =
     styled Html.span
@@ -52,7 +75,15 @@ bottomInfo config =
                             [ String.fromInt charCounter.current ++ "/" ++ String.fromInt charCounter.max |> text ]
                     )
     in
-    [ div [ css [ displayFlex, flexBasis (pct 100), justifyContent spaceBetween, marginTop (rem 0.5), Css.property "gap" "1rem" ] ]
+    [ div
+        [ css
+            [ displayFlex
+            , flexBasis (pct 100)
+            , justifyContent spaceBetween
+            , marginTop (rem 0.5)
+            , Css.property "gap" "1rem"
+            ]
+        ]
         [ config.errorMessage
             |> Html.viewMaybe
                 (\errText ->
--- a/src/Nordea/Components/Common.elm
+++ b/src/Nordea/Components/Common.elm
@@ -1,0 +1,130 @@
+module Nordea.Components.Common exposing (..)
+
+import Css exposing (alignItems, auto, center, color, displayFlex, flex, flexBasis, fontFamilies, fontSize, justifyContent, lineHeight, marginBottom, marginLeft, marginRight, marginTop, none, pct, rem, spaceBetween)
+import Html.Styled as Html exposing (Attribute, Html, div, styled, text)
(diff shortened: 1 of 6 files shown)


src/Nordea/Components/Common.elm
```diff
@@ -0,0 +1,161 @@
...
+                            [ String.fromInt charCounter.current ++ "/" ++ String.fromInt charCounter.max |> text ]
+                    )
+    in
+    [ div
+        [ css
+            [ displayFlex
+            , flexBasis (pct 100)
+            , justifyContent spaceBetween
+            , marginTop (rem 0.5)
+            , Css.property "gap" "1rem"
```
mathilwa, line 84: Can `(rem 1)` be used insted of `"1rem"`?

src/Nordea/Components/Common.elm
```diff
@@ -0,0 +1,161 @@
...
+                            ]
+                        ]
+                        [ Icons.error [ css [ marginRight (rem 0.5), flex none ] ]
+                        , text errText
+                        ]
+                )
+        , charCounterView
+        ]
+        |> showIf (Maybe.isJust config.errorMessage)
+    , div [ css [ displayFlex, flexBasis (pct 100), justifyContent spaceBetween, marginTop (rem 0.5), Css.property "gap" "1rem" ] ]
```
mathilwa, line 105: same here as the comment above

src/Nordea/Components/Label.elm
```diff
@@ -113,6 +89,31 @@ init labelText labelType =
...
+                            Just (Common.Optional a)
+
+                        Just (Custom a) ->
+                            Just (Common.Custom a)
+
+                        Nothing ->
+                            Nothing
+            in
+            { labelText = config.labelText
+            , requirednessHint = config.requirednessHint |> fromLabelRequirednessHint
```
mathilwa, line 110: just wondering about the wording here, requirednessHint. Would requiredHint be sufficient?

src/Nordea/Components/Common.elm
hhaga: I do not think so. As I understood it, the "gap" property is only available through this Css.property and it takes two strings as parameters and need to be written in same way as in CSS.

src/Nordea/Components/Label.elm
hhaga: I think this is a hint regarding the requiredness of the field. I guess requiredHint would work fine as well, but maybe that was not chosen so it should not be mistaken for being a hint that is required (?) 🤔   We actually copied that type (and wording) from `Label` 😅

fredrikloberg (commented):
I'm wondering whether this improves the readability, the only difference I see is the `Label.GroupLabel` param 🤔


```elm
RadioGroup.init
    "Group label"
    |> RadioGroup.withErrorMessage "Some error message"
    |> RadioGroup.view
        []
        [ RadioButton.init
            "simple"
            (text "Click me: ")
            NoOp
            |> RadioButton.withHasError True
            |> RadioButton.view []
        , RadioButton.init
            "simple"
            (text "Click me: ")
            NoOp
            |> RadioButton.withHasError True
            |> RadioButton.view []
         ]
```
Vs.
```elm
Label.init
   "Group label"
    Label.GroupLabel
    |> Label.withErrorMessage "Some error message"
    |> Label.view
        []
        [ RadioButton.init
            "simple"
            (text "Click me: ")
            NoOp
            |> RadioButton.withHasError True
            |> RadioButton.view []
        , RadioButton.init
            "simple"
            (text "Click me: ")
            NoOp
            |> RadioButton.withHasError True
            |> RadioButton.view []
        ]
```

What do you guys think? 🙂

src/Nordea/Components/Common.elm
mathilwa: 👍

src/Nordea/Components/Label.elm
mathilwa: 😅

mathilwa:
> I'm wondering whether this improves the readability, the only difference I see is the `Label.GroupLabel` param 🤔
> 
> ```elm
> RadioGroup.init
>     "Group label"
>     |> RadioGroup.withErrorMessage "Some error message"
>     |> RadioGroup.view
>         []
>         [ RadioButton.init
>             "simple"
>             (text "Click me: ")
>             NoOp
>             |> RadioButton.withHasError True
>             |> RadioButton.view []
>         , RadioButton.init
>             "simple"
>             (text "Click me: ")
>             NoOp
>             |> RadioButton.withHasError True
>             |> RadioButton.view []
>          ]
> ```
> 
> Vs.
> 
> ```elm
> Label.init
>    "Group label"
>     Label.GroupLabel
>     |> Label.withErrorMessage "Some error message"
>     |> Label.view
>         []
>         [ RadioButton.init
>             "simple"
>             (text "Click me: ")
>             NoOp
>             |> RadioButton.withHasError True
>             |> RadioButton.view []
>         , RadioButton.init
>             "simple"
>             (text "Click me: ")
>             NoOp
>             |> RadioButton.withHasError True
>             |> RadioButton.view []
>         ]
> ```
> 
> What do you guys think? 🙂

I think that the biggest improvement in readability here is that its  (in my head) more logical to use radiogroup than label? 😊

mathilwa (approved):

PR #62, opened 2021-11-16
Add  FeatureBox

Rostislav013:
In our machines can't see styles well
image


hhaga (approved):
🎉

src/Nordea/Components/FeatureBox.elm
```diff
@@ -0,0 +1,172 @@
...
+import Nordea.Components.Text as Text exposing (..)
+import Nordea.Html as Html exposing (showIf)
+import Nordea.Resources.Colors as Colors
+import Nordea.Resources.Icons as Icons
+import Nordea.Themes as Themes
+import Svg.Styled as Svg exposing (Svg)
+
+
+type alias Config msg =
+    { showModal : Bool
```
mathilwa, line 66: we now have another component that is called modal - to not confuse them, should we call this showFeatureBox?

src/Nordea/Components/FeatureBox.elm
```diff
@@ -0,0 +1,172 @@
...
+                , overflow auto
+                , displayFlex
+                , justifyContent center
+                , alignItems center
+                , padding4 (rem 8) (rem 1) (rem 2) (rem 1)
+                , backgroundColor (Colors.black |> Colors.withAlpha 0.5)
+                , zIndex (int 1)
+                ]
+            ]
+            [ div
```
mathilwa, line 111: I think you can use the the Card component as the frame for this modal, since it has the same shape 😊

src/Nordea/Components/FeatureBox.elm
```diff
@@ -0,0 +1,172 @@
...
+                [ css
+                    [ displayFlex
+                    , alignItems center
+                    , flexDirection column
+                    , backgroundColor Colors.white
+                    , Css.borderRadius (rem 0.5)
```
mathilwa, line 117: remove Css here and import directly

src/Nordea/Components/FeatureBox.elm
```diff
@@ -0,0 +1,172 @@
...
+                            [ pseudoClass "not(:last-child)" [ marginBottom (rem 0.5) ] ]
+                        ]
+                    , padding (rem 1.5)
+                    ]
+                ]
+                [ closeButton config.closeMsg
+                , config.icon
+                    |> Html.viewMaybe
+                        (\icon ->
+                            div [ css [ marginBottom (rem 1.5) |> Css.important ] ] [ icon ]
```
mathilwa, line 130: Css.important should not be used i think 😅

mathilwa (changes requested):

src/Nordea/Components/FeatureBox.elm
anisha-deverashetty: Yes changed to showFeatureBox 👍

src/Nordea/Components/FeatureBox.elm
anisha-deverashetty: Using Card now

src/Nordea/Components/FeatureBox.elm
anisha-deverashetty: Removed

src/Nordea/Components/FeatureBox.elm
anisha-deverashetty: Yes true. Fixed it now

mathilwa (approved):

PR #65, opened 2021-11-16
Add tooltip component

fshauge:
image


src/Nordea/Components/Tooltip.elm
```diff
@@ -0,0 +1,161 @@
...
+                , Css.top (pct 50)
+                , Css.transform (Css.translate2 (pct 100) (pct -50))
+                ]
+
+        Right ->
+            Css.batch
+                [ Css.left (rem -1)
+                , Css.top (pct 50)
+                , Css.transform (Css.translate2 (pct -100) (pct -50))
+                ]
```
fredrikloberg, line 129: I think we should change the meaning of the placement option so that `Top` means the tooltip is above the content and the arrow pointing down

fredrikloberg (approved):
Nice work 💯

src/Nordea/Components/Tooltip.elm
fshauge: I actually agree, it's just that the Figma sketches are using placement this way. Have to change those as well 😄

PR #69, opened 2021-11-17
Add progress bar and progress bar stepper

fredrikloberg:
Added progress bar and progress bar stepper

Screenshot 2021-11-17 at 16 38 38
Screenshot 2021-11-17 at 16 38 50
Screenshot 2021-11-17 at 16 39 19

### DIFF
--- a/src/Nordea/Components/ProgressBarStepper.elm
+++ b/src/Nordea/Components/ProgressBarStepper.elm
@@ -1,0 +1,84 @@
+module Nordea.Components.ProgressBarStepper exposing (init, view)
+
+import Css
+    exposing
+        ( alignItems
+        , center
+        , color
+        , column
+        , displayFlex
+        , flexDirection
+        , height
+        , marginRight
+        , rem
+        , width
+        )
+import Html.Styled as Html exposing (Attribute, Html)
+import Html.Styled.Attributes exposing (css)
+import List.Extra as List
+import Nordea.Components.ProgressBar as ProgressBar
+import Nordea.Components.Text as Text
+import Nordea.Html exposing (viewMaybe)
+import Nordea.Resources.Colors as Colors
+
+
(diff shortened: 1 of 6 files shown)


mathilwa:
Could this be a part of the already existing stepper? and have two views?

fshauge:
I agree with Mathilde. If we could use media queries to display one or the other or just have an option to choose which one we want.

fredrikloberg:
> Could this be a part of the already existing stepper? and have two views?

Certainly could, for example like we do with buttons. For PartnerHub, we use a custom step indicator for desktop, and the ProgressBarStepper for mobile, so combining them wouldn't yield any benefits for our use.

One thing to note though, is that this ProgressBarStepper and the StepIndicator already in the library work slightly differently. It has one extra parameter, and it's not just a different css styling like it is with button, but a whole new view. Do you think putting them together would make it harder to maintain somehow? 🤔

src/Nordea/Components/ProgressBarStepper.elm
```diff
@@ -0,0 +1,84 @@
...
+                        Text.textTiny
+                            |> Text.view
+                                [ css [ color Colors.grayDark ] ]
+                                [ Html.text (nextLabel strings.next ++ step) ]
+                    )
+            ]
+        ]
+
+
+strings =
```
Eivinddh, line 78: Maybe consider calling it l18n or translation? ☺️  
```suggestion
l18n =
```

fredrikloberg: We have used `strings` for this throughout in PartnerHub, which is also the same name used for this in native Android development. `strings` is already used in the component library as well.

fredrikloberg:
I will merge this now, but feel free to refactor if you think the changes it improve it 🙂

PR #70, opened 2021-11-19
Button-Card-fix: Changed hover effect. Added examples of how button c…

vetlesolgaard:
Fixes some properties on the `Button.card`

Screenshots of examples👇
The upper one is displaying the backgroundcolor that appears when it is hovered over.
Screenshot 2021-11-19 at 14 29 12

### DIFF
--- a/src/Nordea/Components/Button.elm
+++ b/src/Nordea/Components/Button.elm
@@ -7,6 +7,7 @@ module Nordea.Components.Button exposing
     , secondary
     , tertiary
     , view
+    , withHtmlTag
     , withStyles
     )
 
@@ -30,6 +31,7 @@ import Css
         , fontFamilies
         , fontSize
         , fontWeight
+        , height
         , hover
         , int
         , left
@@ -42,13 +44,19 @@ import Css
         , pointer
         , pointerEvents
         , rem
-        , scale
         , solid
+        , start
         , textAlign
-        , transform
+        , transforms
+        , translateX
+        , width
         )
+import Css.Global exposing (children, descendants, everything, withClass)
+import Css.Transitions exposing (easeOut, transition)
 import Html.Styled as Html exposing (Attribute, Html)
+import Html.Styled.Attributes exposing (class, css)
 import Nordea.Resources.Colors as Colors
+import Nordea.Resources.Icons as Icons
 import Nordea.Themes as Themes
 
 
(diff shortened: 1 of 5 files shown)


src/Nordea/Components/Text.elm
```diff
@@ -211,7 +221,6 @@ variantStyle variant =
                 , fontWeight (int 500)
                 , fontSize (rem 3)
                 , lineHeight (rem 3.5)
-                , margin zero
```
vetlesolgaard, line 214: Any reason all these `margin zero` must be set for the text styles? 🤔  @fredrikloberg

fredrikloberg: Generic components like these dont know how and where they are going to be used, so having a default margin zero makes them easier to use.`h` and `p` tags have all kinds of margins by default

fredrikloberg: The idea is to add the margins you need when you use the component

vetlesolgaard: Aah, okay. In the cases we want to add a marging we'll have to use `|> Css.important`. Do you know if there is any way to say that margin is zero unless anything else is specified? So we won't have to use the `important` style? 😊

vetlesolgaard: I guess the `Html.span` don't need the `margin zero` property

fredrikloberg: Ye, overriding styles is a problem with elm-css :/ Maybe we can set the zero margins with a less specific css selector

fredrikloberg: The span elements dont need it, but we thought it was nice to just add it to every variant because the actual html tag may be changed depending on the context

src/Nordea/Components/Button.elm
```diff
@@ -108,13 +112,22 @@ withStyles styles (Button config) =
...
 view : List (Attribute msg) -> List (Html msg) -> Button -> Html msg
 view attributes children (Button config) =
+    let
+        variantSpecificChildren =
+            case config.variant of
+                Card ->
+                    [ Icons.arrowRight [ css [ width (rem 1), height (rem 1) ] ] ]
+
+                _ ->
+                    [ Html.nothing ]
```
fredrikloberg, line 122:
```suggestion
                    []
```

src/Nordea/Components/Button.elm
```diff
@@ -108,13 +112,22 @@ withStyles styles (Button config) =
...
+    in
     Html.styled Html.button
```
fredrikloberg, line 124: The front page of PH needs these card buttons as a link instead of a button to be semantically correct. Based on the answer [here](https://stackoverflow.com/questions/6393827/can-i-nest-a-button-element-inside-an-a-using-html5), I see that we cannot simply wrap these buttons in a `a`-tag where needed, as that is not valid html.

I think we need a way to change the tag dynamically. Maybe we can do it the same way we do it for the typography?

src/Nordea/Resources/Colors.elm
```diff
@@ -118,6 +119,11 @@ grayWarm =
     hex "#F4F2F1"
 
 
+hoverGray : Color
+hoverGray =
```
fredrikloberg, line 123: The naming pattern for the other colors is to have the color first and a variant second:
```suggestion
grayHover : Color
grayHover =
```

It makes autocomplete better 🤩

src/Nordea/Components/Button.elm
```diff
@@ -63,58 +71,65 @@ type Variant
...
 
-view : List (Attribute msg) -> List (Html msg) -> Button -> Html msg
+view : List (Attribute msg) -> List (Html msg) -> Button msg -> Html msg
 view attributes children (Button config) =
-    Html.styled Html.button
+    let
+        variantSpecificChildren =
+            case config.variant of
+                Card ->
+                    [ Icons.arrowRight [ attribute "type" "arrowicon", css [ width (rem 1), height (rem 1) ] ] ]
```
fredrikloberg, line 121: Is the `arrowicon` type attribute only used to to apply the movement on hover? If that is the case, I think it is more correct to use a classname for that. Based on [this](https://www.geeksforgeeks.org/html-type-attribute/), it doesnt look like the type is valid on div elements.

```suggestion
                    [ Icons.arrowRight [ class "arrowicon", css [ width (rem 1), height (rem 1) ] ] ]
```

src/Nordea/Components/Button.elm
```diff
@@ -196,17 +211,45 @@ variantStyle variant =
...
+-- EXTRAS
+
+
+withStyles : List Style -> Button msg -> Button msg
+withStyles styles (Button config) =
+    Button { config | styles = styles }
+
+
+withHtmlTagWrapper : (List (Attribute msg) -> List (Html msg) -> Html msg) -> Button msg -> Button msg
+withHtmlTagWrapper htmlTag (Button config) =
```
fredrikloberg, line 254: Just nitpicking 😛 As the tag isn't wrapping the component, but the tag of the component itself, I think we should avoid wrapper in the name

```suggestion
withHtmlTag : (List (Attribute msg) -> List (Html msg) -> Html msg) -> Button msg -> Button msg
withHtmlTag htmlTag (Button config) =
```

fredrikloberg (approved):
Looks good!

PR #72, opened 2021-11-23
Button-card-additional-PAR-673: Updated needed properties

vetlesolgaard:
Added needed properties to work smoother when using the `Button.card` component


hungqt (approved):
Nice 🚀

PR #73, opened 2021-11-24
Add simple dropdown variant

fredrikloberg:
In PartnerHub, we need a simple variant of the dropdown to put in the header.

Screenshot 2021-11-24 at 12 04 04


vetlesolgaard (approved):
🙌

PR #74, opened 2021-11-25
Change how to use attrs inside Card

hungqt:
Change how to use attrs inside Card + small fixes

**Example reason for change:**
In PartnerHub we use conditional rendering some places to decide if we want to show the Card component or not. Using `attrs` in the root-component here would be nice instead of `attrs` being used in the child-component. Any thoughts on why this _shoulndt_ be done like this?


fredrikloberg (approved):
Great!

PR #75, opened 2021-11-25
Fix: Flex column in Card component 

hungqt:
Fixed the Card-component. I Added flex column to the wrong component last PR 😬

### DIFF
--- a/src/Nordea/Components/Card.elm
+++ b/src/Nordea/Components/Card.elm
@@ -1,22 +1,6 @@
 module Nordea.Components.Card exposing (..)
 
-import Css
-    exposing
-        ( Style
-        , backgroundColor
-        , borderBottom3
-        , borderRadius
-        , column
-        , displayFlex
-        , flexDirection
-        , fontSize
-        , left
-        , padding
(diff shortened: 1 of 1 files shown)


src/Nordea/Components/Card.elm
```diff
@@ -47,9 +47,7 @@ view : List (Attribute msg) -> List (Html msg) -> Card msg -> Html msg
 view attrs children (Card config) =
     let
         cardStyle =
-            [ displayFlex
-            , flexDirection column
-            , borderRadius (rem 0.5)
+            [ borderRadius (rem 0.5)
```
fredrikloberg, line 50: Hmm, I suggest we remove the `cardContentContainer` instead as that would make it easier to use. 

Adding padding to this, and `marginBottom (rem 1)` to the `cardTitle` (instead of the padding) does the trick I think
```suggestion
            [ displayFlex
            , flexDirection column
            , borderRadius (rem 0.5)
            , padding (rem 2)
```

fredrikloberg (approved):

PR #78, opened 2021-11-29
Add components for error pages

mathilwa:
Add components for showing two type of errors, internal server error and page not found.

Internal server error has fixed texts, but its possible to send in children to add custom information for your use. 

Screen Shot 2021-11-29 at 12 23 06 PM


src/Nordea/Components/Error.elm
```diff
@@ -0,0 +1,222 @@
...
+init : ErrorConfig -> Error msg
+init config =
+    Error { locale = config.locale, errorType = config.errorType }
+
+
+view : List (Attribute msg) -> List (Html msg) -> Error msg -> Html msg
+view attributes children (Error config) =
+    let
+        toI18NString textToTranslate =
+            config.locale textToTranslate
```
fshauge, line 50: Not sure I get this... 😅 Isn't this the same as just using `config.locale`? You define a function that just calls another function. If you want to rename the variable, you could at least do something like this:

```elm
toI18NString =
    config.locale
```

Right?

src/Nordea/Components/Error.elm
```diff
@@ -0,0 +1,222 @@
...
+            ]
+        ]
+        [ Html.div [ Attributes.css [ Css.marginBottom (rem 1) ] ]
+            [ Text.headlineTwo
+                |> Text.withHtmlTag Html.h1
+                |> Text.view
+                    [ Attributes.css [ Css.textAlign Css.center ]
+                    ]
+                    [ Html.text (texts.heading |> toI18NString) ]
+            ]
```
fredrikloberg, line 84:
```suggestion
        [ Text.headlineTwo
                |> Text.withHtmlTag Html.h1
                |> Text.view
                    [ Attributes.css [ Css.textAlign Css.center, Css.marginBottom (rem 1) ] ]
                    [ Html.text (texts.heading |> toI18NString) ]
```

src/Nordea/Components/Error.elm
```diff
@@ -0,0 +1,222 @@
...
+                [ Css.marginBottom (rem 2)
+                , Css.maxWidth (rem 30)
+                , Css.textAlign Css.center
+                ]
+            ]
+            [ Text.bodyTextLight
+                |> Text.view []
+                    [ Html.text errorDescription
+                    ]
+            ]
```
fredrikloberg, line 96:
```suggestion
        , Text.bodyTextLight
                |> Text.view 
                     [ Attributes.css
                         [ Css.marginBottom (rem 2)
                         , Css.maxWidth (rem 30)
                         , Css.textAlign Css.center -- Not sure if this is needed since the wrapping div as alignItems center
                         ]
                    ]
                    [ Html.text errorDescription ]
```

src/Nordea/Components/Error.elm
```diff
@@ -0,0 +1,222 @@
...
+        , Html.div [] [ errorSvg ]
+        , showIf
+            (config.errorType
+                == InternalServerError
+            )
+            (viewActionForInternalServerError errorAction)
```
fredrikloberg, line 102: Just a nitpick/personal preference

```suggestion
        , viewActionForInternalServerError errorAction
            |> showIf (config.errorType == InternalServerError)
```

src/Nordea/Components/Error.elm
```diff
@@ -0,0 +1,222 @@
...
+        , Html.div
+            (Attributes.css [ Css.maxWidth (rem 30) ] :: attributes)
+            children
+        ]
```
fredrikloberg, line 106: I think this makes it easier to use and control, as css you may provide to `Error.view []` will then also affect the supplied children.

```suggestion
        ] ++ children
```

src/Nordea/Components/Error.elm
mathilwa: You are completely right. Fixed it now!

src/Nordea/Components/Error.elm
```diff
@@ -0,0 +1,222 @@
...
+            , Css.maxWidth (rem 30)
+            , Css.alignItems Css.center
+            ]
+        ]
+        [ Text.bodyTextLight
+            |> Text.view [ Attributes.css [ Css.paddingRight (rem 0.25), Css.textAlign Css.center ] ]
+                [ Html.text actionText
+                , FlatLink.default |> FlatLink.view [ href "mailto: kundeservice.nfe@nordea.com", Attributes.css [ Css.display Css.inlineBlock ] ] [ Html.text "kundeservice.nfe@nordea.com." ]
+                ]
+        ]
```
fredrikloberg, line 124:
```suggestion
viewActionForInternalServerError : String -> Html msg
viewActionForInternalServerError actionText =
    Text.bodyTextLight
            |> Text.view 
                [ Attributes.css 
                    [ Css.paddingRight (rem 0.25)
                    , Css.textAlign Css.center 
                    , Css.displayFlex
                    , Css.flexDirection Css.column
                    , Css.maxWidth (rem 30)
                    , Css.alignItems Css.center
                    ] 
                 ]
                [ Html.text actionText
                , FlatLink.default |> FlatLink.view [ href "mailto: kundeservice.nfe@nordea.com", Attributes.css [ Css.display Css.inlineBlock ] ] [ Html.text "kundeservice.nfe@nordea.com." ]
                ]
```

src/Nordea/Components/Error.elm
```diff
@@ -0,0 +1,222 @@
...
+init : ErrorConfig -> Error msg
+init config =
+    Error { locale = config.locale, errorType = config.errorType }
+
+
+view : List (Attribute msg) -> List (Html msg) -> Error msg -> Html msg
+view attributes children (Error config) =
+    let
+        toI18NString =
+            config.locale
```
fredrikloberg, line 50: Would it be better to rename the property on the config instead? Maybe to `translate`?

src/Nordea/Components/Error.elm
```diff
@@ -0,0 +1,222 @@
...
+
+            else
+                texts.internalServerError.description |> toI18NString
+
+        errorAction =
+            if config.errorType == InternalServerError then
+                texts.internalServerError.action |> toI18NString
+
+            else
+                ""
```
fredrikloberg, line 64: I'd say pass the translate function to `viewActionForInternalServerError` directly and let that function pick the right text

fredrikloberg (approved):
Looks good! 🙌 Just some nitpicky comments

src/Nordea/Components/Error.elm
fshauge: 👆

src/Nordea/Components/Error.elm
mathilwa: Actually, that will not work because the text component has margin: 0 that overrides that marginBottom styling. I had the styles there before. Thats why I added the outer div

src/Nordea/Components/Error.elm
mathilwa: The wrapping div just align the body text, but the text is not centered if it e.g its shown on two or more lines. So it need textAlign as well

src/Nordea/Components/Error.elm
mathilwa: from locale to translate you mean?

fshauge: Jupp

src/Nordea/Components/Error.elm
fredrikloberg: I'd prefer a Css.important over a wrapping div in that case. But yea, maybe we can find a better solution for margin in the text component 🤔

mathilwa: I would prefer not to use Css.important, we should not be using that too much 😅

src/Nordea/Components/Error.elm
mathilwa: 👍

src/Nordea/Components/Error.elm
mathilwa: The width of extra information should not exceed 30 rem. Do you still want the users of the component to control that themselves?

src/Nordea/Components/Error.elm
mathilwa: The same comment her as the one further up. The text component clears the marginBottom because it has margin: 0

src/Nordea/Components/Error.elm
mathilwa: 👍

src/Nordea/Components/Error.elm
```diff
@@ -0,0 +1,219 @@
...
+init : ErrorConfig -> Error msg
+init config =
+    Error { translate = config.translate, errorType = config.errorType }
+
+
+view : List (Attribute msg) -> List (Html msg) -> Error msg -> Html msg
+view attributes children (Error config) =
+    let
+        toI18NString =
+            config.translate
```
fredrikloberg, line 50: I think using `config.translate` directly is fine 
```suggestion
```

src/Nordea/Components/Error.elm
fredrikloberg: But does the div have a margin in this case? 🙂

src/Nordea/Components/Error.elm
fredrikloberg: Ye I agree that overusing Css.important is usually a code smell, but in cases like these, I think it conveys the problem more clearly. We add Css.important to _override_ something. I find the intent of a wrapping div harder to infer.

The underlying problem is really in the elm-css library though, I really wish it preserved the order of the styles and not put them in random order 😕

Anyway, just my thoughts, go with whichever solution you think is best.

src/Nordea/Components/Error.elm
mathilwa: 👍

PR #81, opened 2021-11-30
Add new init function to allow disabled options without breaking backwards

hannabergljot:
Dropdown component can be initiated with only value and text using the init function as before. For Dropdown components with disabled options, initWithOptionsProperties can be used:)

### DIFF
--- a/src/Nordea/Components/Dropdown.elm
+++ b/src/Nordea/Components/Dropdown.elm
@@ -100,19 +100,7 @@ simple options optionToString onInput =
 
 init : List { value : a, text : String } -> (a -> String) -> (a -> msg) -> Dropdown a msg
 init options optionToString onInput =
-    let
-        optionsWithDisabled =
-            List.map (\option -> optionInit option) options
-    in
-    Dropdown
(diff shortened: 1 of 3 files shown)


src/Nordea/Components/Dropdown.elm
```diff
@@ -98,8 +98,25 @@ simple options optionToString onInput =
...
+    in
+    Dropdown
+        { placeholder = Nothing
+        , onInput = onInput
+        , options = optionsWithDisabled
+        , optionToString = optionToString
+        , selectedValue = Nothing
+        , hasError = False
+        , variant = Standard
+        }
```
fredrikloberg, line 115:
```suggestion
    initWithOptionProperties (List.map optionInit options) optionToString onInput
```

fredrikloberg (approved):
Nice solution 🤩

src/Nordea/Components/Dropdown.elm
hannabergljot: Clever😎

mathilwa (approved):
Nice!

PR #82, opened 2021-11-30
Add info panel to list information

mathilwa:
New info panel component. It lists information together with label in a grey box, separated with borders.

### DIFF
--- a/src/Nordea/Components/InfoPanel.elm
+++ b/src/Nordea/Components/InfoPanel.elm
@@ -16,11 +16,11 @@ view : List InfoPanelContent -> Html msg
 view panelInfoList =
     Html.div
         [ Attributes.css
-            [ Css.backgroundColor Colors.grayCool
+            [ Css.backgroundColor (Colors.grayCool |> Colors.withAlpha 0.5)
             , Css.maxWidth (rem 25)
             , Css.boxSizing Css.borderBox
             , Css.borderRadius (rem 0.5)
-            , Css.border3 (rem 0.125) Css.solid Colors.grayMedium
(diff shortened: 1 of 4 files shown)


Rostislav013 (approved):
👍

src/Nordea/Components/InfoPanel.elm
```diff
@@ -0,0 +1,47 @@
...
+        ]
+        [ Html.div
+            [ Attributes.css
+                [ Css.fontSize (rem 0.875)
+                , Css.fontWeight (int 200)
+                , Css.marginBottom (rem 0.25)
+                ]
+            ]
+            [ Html.text infoPanelContent.label ]
+        , Html.text infoPanelContent.info
```
fredrikloberg, line 46: Can the Text component be used here?

mathilwa: It might be used for the label? But for the info text it will change both font size and weight, so I dont know if that will just complicate the work? 🤷

fredrikloberg: Hmm, I think we have a "bold" variant of each text, but is the font size something custom?

mathilwa: For the first information listed, the font size should be bigger and also bold. For the others its just 1 rem. and the font weight should also be regular. Since there is more than one adjustment, I went for regular text instead of trying to make our text components fit this solution.

But I am open for suggestions! 😊

fredrikloberg: I think this approach is fine, but maybe specify the Nordea font explicitly? Just in case the users of the component havent defined it globally.

fredrikloberg (approved):

PR #83, opened 2021-11-30
Fix the use of :first-child css selector in UI Explorer

fredrikloberg:
The style tag from elm-css is added as the first child of the root component (from the Html.Styled package). When a component is the root and using the first-child css selector, the style tag from elm-css becomes the first child, which breaks the styling of the component.


fshauge (approved):
Disappointing that elm-css uses this approach in the first place as it breaks CSS

PR #84, opened 2021-11-30
Update border radius, height, and hover effect

fredrikloberg:
Update border radius, height, and hover effect of some components


fshauge (approved):

mathilwa (approved):

hhaga (approved):
💯

PR #85, opened 2021-12-01
Add-table-component: Added basic structure for table

vetlesolgaard:
- Adding a first edition of a table component of type A table (See screenshot)
- This first edition will miss a lot of the features that is displayed in the screenshot, but implements the basic building structures

Screenshot 2021-12-01 at 15 16 31

How it looks when showcased in the component library 👇
Screenshot 2021-12-01 at 15 17 48

### DIFF
--- a/src/Nordea/Components/Table.elm
+++ b/src/Nordea/Components/Table.elm
@@ -9,28 +9,34 @@ module Nordea.Components.Table exposing
 
 import Css
     exposing
-        ( backgroundColor
+        ( alignSelf
+        , backgroundColor
         , borderBottom3
+        , center
         , column
         , displayFlex
+        , ellipsis
         , flexDirection
+        , height
+        , hidden
         , hover
-        , left
+        , noWrap
         , nthChild
-        , paddingBottom
-        , paddingLeft
-        , paddingTop
-        , px
+        , overflow
+        , padding
+        , padding4
         , rem
         , row
         , solid
-        , textAlign
+        , textOverflow
+        , whiteSpace
         )
-import Css.Global exposing (descendants, typeSelector)
+import Css.Global as Css exposing (typeSelector)
 import Html.Styled as Html exposing (Attribute, Html)
 import Html.Styled.Attributes exposing (css)
 import Nordea.Components.Text as Text
 import Nordea.Resources.Colors as Colors
+import Nordea.Themes as Themes
 
 
 view : List (Attribute msg) -> List (Html msg) -> Html msg
@@ -46,7 +52,13 @@ view attrs children =
 -}
 thead : List (Attribute msg) -> List (Html msg) -> Html msg
 thead attrs children =
-    Html.thead (css [ borderBottom3 (px 1) solid Colors.grayLight ] :: attrs)
+    Html.thead
+        (css
+            [ height (rem 3.0)
+            , borderBottom3 (rem 0.0625) solid Colors.grayLight
+            ]
+            :: attrs
+        )
         children
 
 
@@ -58,7 +70,6 @@ tr attrs children =
         (css
             [ displayFlex
             , flexDirection row
-            , hover [ backgroundColor Colors.blueCloud |> Css.important ]
             ]
             :: attrs
         )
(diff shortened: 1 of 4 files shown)


src/Nordea/Components/Table.elm
```diff
@@ -0,0 +1,111 @@
...
+
+
+{-| Groups the header content in a table
+-}
+thead : List (Attribute msg) -> List (Html msg) -> Html msg
+thead attrs children =
+    Html.thead
+        (css
+            [ height (rem 3.0)
+            , borderBottom3 (px 1) solid Colors.grayLight
```
fredrikloberg, line 52:
```suggestion
            , borderBottom3 (rem 0.0625) solid Colors.grayLight
```

src/Nordea/Components/Table.elm
```diff
@@ -0,0 +1,111 @@
...
+        children
+
+
+{-| Defines a header cell in a table
+-}
+th : List (Attribute msg) -> List (Html msg) -> Html msg
+th attrs children =
+    Html.th
+        (css [ padding4 (rem 1.0) (rem 0.75) (rem 0.75) (rem 0.75) ] :: attrs)
+        [ Text.init Text.TextSmallHeavy
```
fredrikloberg, line 79:
```suggestion
        [ Text.textSmallHeavy
```

src/Nordea/Components/Table.elm
```diff
@@ -0,0 +1,111 @@
...
+tbody : List (Attribute msg) -> List (Html msg) -> Html msg
+tbody attrs children =
+    Html.tbody
+        (css
+            [ displayFlex
+            , flexDirection column
+            , descendants
+                [ typeSelector "tr"
+                    [ height (rem 4.5)
+                    , hover [ backgroundColor Colors.blueCloud |> Css.important ]
```
fredrikloberg, line 95: We must remember to use theme colors for most blue colors 🙂 Also, is the Css.important needed?


```suggestion
                    , hover [ Themes.backgroundColor Themes.SecondaryColor Colors.blueCloud ]
```

src/Nordea/Components/Table.elm
```diff
@@ -0,0 +1,111 @@
...
+-}
+tbody : List (Attribute msg) -> List (Html msg) -> Html msg
+tbody attrs children =
+    Html.tbody
+        (css
+            [ displayFlex
+            , flexDirection column
+            , descendants
+                [ typeSelector "tr"
+                    [ height (rem 4.5)
```
fredrikloberg, line 94: Do we need to define a height? I'm thinking maybe padding is enough to get a default row height, and then the height property can be specified if needed when the table is used?
```suggestion
                    [ height (rem 4.5)
```

fredrikloberg (approved):
Noice, I really like following the standard Html structure and naming 🤩 💯

src/Nordea/Components/Table.elm
vetlesolgaard: Aah thanks for the reminder 😍
Yeah, since we're coloring every other row, these already have a background color, so to override this I thiiink the `Css.important` is necessary

fredrikloberg: Just a guess, but maybe changing `descendants` to `children` gives this more specificity and enough to override the alternate color without important

src/Nordea/Components/Table.elm
vetlesolgaard: Prooobably not. Just seemed like that height was default on all the examples. If the row height isn't set here, then we can't set default padding either, since it will affect text and icons inside the rows differently 🤔 
I can ask Thea if the table rows should have a set height or not 😊

vetlesolgaard: Prooobably not. Just seemed like that height was default on all the examples. If the row height isn't set here, then we can't set default padding either, since it will affect text and icons inside the rows differently 🤔
I can ask Thea if the table rows should have a set height or not 😊

src/Nordea/Components/Table.elm
vetlesolgaard: Aaah, I never really checked out the difference between the two. I'll try 🙌

src/Nordea/Components/Table.elm
fredrikloberg: Hmm, will padding work if we add `alignItems center`? 🤔

vetlesolgaard: No, looks like it will still shrink to fit whatever content there is + padding 😬

src/Nordea/Components/Table.elm
vetlesolgaard: Tried changing to `children`, looks like `Css.important` is still needed 👎

fshauge:
Alignment looks to be off

image

vetlesolgaard:
> Alignment looks to be off
> 
> image

Ah! It is just examples of how it can be used. Maybe a bit misleading 😅 Should I just center fix and center all texts?

PR #88, opened 2021-12-06
Add search icon to text input

hungqt:
Added search icon to TextInput with `withSearchIcon`

**NB: Had to add a div outside of the input element itself in order to do this. Will this change affect anything in the various applications (Can affect text inputs with styling by child attributes)**

image


src/Nordea/Components/TextInput.elm
```diff
@@ -92,16 +111,36 @@ withError condition (TextInput config) =
...
+                [ position absolute
+                , left (rem 0.7)
+                , top (pct 50)
+                , transform (translateY (pct -50))
+                , displayFlex
+                , alignItems center
+                , pointerEvents none
+                ]
+            ]
+            [ Icons.search [ css [ width (rem 1.4), opacity (num 0.5) ] ] ]
```
fredrikloberg, line 137: Can the span be avoided by adding the styled directly to the icon?

```suggestion
        [ Icons.search 
            [ css 
                [ position absolute
                , left (rem 0.7)
                , top (pct 50)
                , transform (translateY (pct -50))
                , displayFlex
                , alignSelf center
                , pointerEvents none
                , width (rem 1.4)
                , opacity (num 0.5) 
                ] 
            ]
```

src/Nordea/Components/TextInput.elm
```diff
@@ -92,16 +111,36 @@ withError condition (TextInput config) =
...
 -- VIEW
 
 
 view : List (Attribute msg) -> TextInput msg -> Html msg
 view attributes (TextInput config) =
-    styled input
-        (getStyles config)
-        (getAttributes config ++ attributes)
-        []
+    Html.div [ css [ position relative ] ]
```
fredrikloberg, line 125:
```suggestion
    Html.div [ css [ displayFlex, position relative ] ]
```

src/Nordea/Components/TextInput.elm
```diff
@@ -92,16 +111,36 @@ withError condition (TextInput config) =
...
+                , displayFlex
+                , alignItems center
+                , pointerEvents none
+                ]
+            ]
+            [ Icons.search [ css [ width (rem 1.4), opacity (num 0.5) ] ] ]
+            |> showIf config.hasSearchIcon
+        , styled input
+            (getStyles config)
+            (getAttributes config ++ attributes)
```
fredrikloberg, line 141: I think the attributes should be added to the root element, in this case the wrapping div, to better support adding margins etc.

```suggestion
            (getAttributes config)
```

fredrikloberg (approved):
Looks good! One idea to avoid the potential breaking change is to only wrap the input in a div if `hasSearchIcon` is set to True

hungqt:
Updated now. Did you mean like this? @fredrikloberg

fredrikloberg:
> Updated now. Did you mean like this? @fredrikloberg

Yeah, shouldn't be breaking now 🚀

src/Nordea/Components/TextInput.elm
```diff
@@ -92,16 +111,46 @@ withError condition (TextInput config) =
...
+                    [ width (rem 1.4)
+                    , opacity (num 0.5)
+                    , position absolute
+                    , left (rem 0.7)
+                    , top (pct 50)
+                    , transform (translateY (pct -50))
+                    , displayFlex
+                    , alignItems center
+                    , pointerEvents none
+                    ]
```
fredrikloberg, line 141: Are these needed now?

```suggestion
        Html.div
            (css [ position relative ]
                :: attributes
            )
            [ Icons.search
                [ css
                    [ width (rem 1.4)
                    , opacity (num 0.5)
                    , position absolute
                    , left (rem 0.7)
                    , top (pct 50)
                    , transform (translateY (pct -50))
                    , pointerEvents none
                    ]
```

hungqt: Nope!

PR #89, opened 2021-12-06
Update typography

fredrikloberg:
Update typography

### DIFF
--- a/src/Nordea/Resources/Fonts.elm
+++ b/src/Nordea/Resources/Fonts.elm
@@ -9,7 +9,7 @@ fromSize : Float -> Style
 fromSize size =
     let
         fontFamily =
-            if size < 1.375 then
+            if size <= 1.375 then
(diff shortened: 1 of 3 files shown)


src/Nordea/Resources/Fonts.elm
```diff
@@ -9,7 +9,7 @@ fromSize : Float -> Style
 fromSize size =
     let
         fontFamily =
-            if size < 1.125 then
+            if size <= 1.375 then
```
fshauge, line 12: Hmm, this seems ambiguous... both fonts can be used for 22px

image

fshauge: But anyway, 1.375rem is more correct than 1.125rem 😅

fshauge: nvm lol

fshauge (approved):
🌶️

PR #90, opened 2021-12-06
Fix dropdown padding

fredrikloberg:
Before: 
Screenshot 2021-12-06 at 15 26 32
After: 
Screenshot 2021-12-06 at 15 26 58


fshauge (approved):
:shipit:

PR #92, opened 2021-12-14
PAR-Add-info-label-component: Added InfoLabel component

vetlesolgaard:
Added `InfoLabel` component

Screenshot 2021-12-14 at 09 29 35
Screenshot 2021-12-14 at 09 29 40


src/Nordea/Components/InfoLabel.elm
```diff
@@ -0,0 +1,56 @@
...
+
+
+viewWarning : List (Attribute msg) -> List (Html msg) -> Html msg
+viewWarning attrs children =
+    Html.div
+        (infoLabelContainerStyle :: css [ backgroundColor Colors.yellow ] :: attrs)
+        [ Icon.info [ css [ width (rem 1.5), height (rem 1.5), marginRight (rem 0.5) ] ]
+        , Text.bodyTextSmall
+            |> Text.view [] children
+        ]
```
vetlesolgaard, line 44: Unsure how to make the use of this optimal. `InfoLabel.viewWarning [] []` doesn't feel optimal, but at the same time I would like to avoid the `Component.componentType |> Component.view [] []` pattern since it feels so verbose. Any thoughts?

src/Nordea/Components/InfoLabel.elm
```diff
@@ -0,0 +1,64 @@
...
+
+infoLabelContainerStyle : Attribute msg
+infoLabelContainerStyle =
+    css
+        [ displayFlex
+        , flexDirection row
+        , padding (rem 0.75)
+        , borderRadius (rem 0.5)
+        , maxWidth fitContent
+        , marginBottom (rem 1)
```
fredrikloberg, line 58: Can we avoid the margin?

```suggestion
```

src/Nordea/Components/InfoLabel.elm
```diff
@@ -0,0 +1,64 @@
...
+        ]
+
+
+labelTextStyle : Attribute msg
+labelTextStyle =
+    css [ lineHeight (rem 0) |> Css.important, alignSelf center ]
```
fredrikloberg, line 64: Do we need to set the lineHeight here?
```suggestion
    css [ alignSelf center ]
```

fredrikloberg (approved):
I think the approach of having multiple views works fine in this case. We can always change it later of the component expands

src/Nordea/Components/InfoLabel.elm
vetlesolgaard: Haha nope. Not needed. Thought it was a little bit off-centered, but the same happens when lineheight is present 🙃 I'll remove it

src/Nordea/Components/InfoLabel.elm
vetlesolgaard: Yeah! Good spot🙌 Was left from when copying over from PartnerHub

src/Nordea/Components/InfoLabel.elm
vetlesolgaard: `alignSelf center` is actually not needed as well 😁

PR #102, opened 2022-01-05
Add TextArea component

fonnes:
Hello!
In this pull request, we add TextArea to elm components library. There is another PR which has begun implementing this, but since the PR has not been active for some time, and I wanted to learn about this library, we decided to create a new PR instead.

Link to figma: https://www.figma.com/file/XSQT0h4JCmpWsCq4UyXWN3/Nordea-Design-Web-toolkit-V1.4.1?node-id=4221%3A33995

Default:
image

Placeholder:
image

Filled:
image

Error:
image

Disabled:
image


src/Nordea/Components/TextArea.elm
```diff
@@ -0,0 +1,126 @@
...
+    let
+        borderColorStyle =
+            if config.showError then
+                Colors.redDark
+
+            else
+                Colors.grayMedium
+    in
+    [ fontSize (rem 1)
+    , padding4 (rem 0.5) (rem 0.75) (rem 0.5) (rem 0.75)
```
fredrikloberg, line 114: Makes sure the correct font is used in case it is not specified globally

```suggestion
    , fontFamilies [ "Nordea Sans Small", "sans-serif" ]
    , padding4 (rem 0.5) (rem 0.75) (rem 0.5) (rem 0.75)
```

fredrikloberg (approved):
Looks good! 💯

In PartnerHub, we also require an `onBlur` event. So would be great if you can also add support for that 🙏

src/Nordea/Components/TextArea.elm
fshauge: Usually when you need the font you should use `Font.fromSize` to get the correct font family based on the font size. We use `Nordea Sans Small` and `Nordea Sans Large` based on the font size. There are more details regarding these under Design Tokens in Figma

fshauge (approved):
Looks very good! 🏅  Make sure to run `elm-format` to tidy up formatting 😉

src/Nordea/Components/TextArea.elm
fshauge: Actually specifying `sans-serif` as a fallback when the font is loaded is a good idea. Right now the fallback is `serif` which doesn't look particularly good 😅 I'll make a PR for this

src/Nordea/Components/TextArea.elm
```diff
@@ -0,0 +1,134 @@
...
+                Colors.redDark
+
+            else
+                Colors.grayMedium
+
+        fontSizeNum =
+            1
+    in
+    [ fontSize (rem fontSizeNum)
+    , Fonts.fromSize fontSizeNum
```
fshauge, line 121: Should have been more precise 😄 `Fonts.fromSize` will give you both `Css.fontFamilies` and `Css.fontSize`, so no need to specify it manually.

```suggestion
    in
    [ Fonts.fromSize 1
```

fonnes: Aha, nice! Will fix :D

src/Nordea/Components/TextArea.elm
```diff
@@ -0,0 +1,136 @@
...
+    let
+        borderColorStyle =
+            if config.showError then
+                Colors.redDark
+
+            else
+                Colors.grayMedium
+    in
+    [ Fonts.fromSize 1
+    , padding4 (rem 0.5) (rem 0.75) (rem 0.5) (rem 0.75)
```
mathilwa, line 124: we can use padding2 here  `padding2 (rem 0.5) (rem 0.75)` as long the top/bottom and left/right padding are the same 😊

PR #105, opened 2022-01-06
Move number input to the right

fonnes:
In this PR we are moving the input of the number input to the right. 
We are also removing the default arrow icon which increases and decreases the input number, because it was colliding with the right input.

image

### DIFF
--- a/src/Nordea/Components/NumberInput.elm
+++ b/src/Nordea/Components/NumberInput.elm
@@ -27,7 +27,9 @@ import Css
         , padding2
         , pct
         , rem
+        , right
         , solid
+        , textAlign
         , width
         )
 import Html.Styled exposing (Attribute, Html, input, styled)
@@ -141,6 +143,7 @@ getStyles config =
     in
     [ fontSize (rem 1)
     , height (rem 3)
+    , textAlign right
     , padding2 (rem 0.75) (rem 0.75)
     , borderRadius (rem 0.25)
     , border3 (rem 0.0625) solid borderColorStyle
--- a/src/Nordea/Components/NumberInput.elm
+++ b/src/Nordea/Components/NumberInput.elm
@@ -19,10 +19,10 @@ import Css
         , borderRadius
         , boxSizing
         , disabled
+        , display
         , focus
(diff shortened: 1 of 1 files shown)


src/Nordea/Components/NumberInput.elm
```diff
@@ -141,6 +147,10 @@ getStyles config =
     in
     [ fontSize (rem 1)
     , height (rem 3)
+    , pseudoElement "-webkit-outer-spin-button" [ property "-webkit-appearance" "none", margin (px 0) ]
+    , pseudoElement "-webkit-inner-spin-button" [ property "-webkit-appearance" "none", margin (px 0) ]
```
fredrikloberg, line 151: Will this work? 🤔
```suggestion
    , pseudoElement "-webkit-outer-spin-button" [ display none ]
    , pseudoElement "-webkit-inner-spin-button" [ display none ]
```

If not then I suggest to change `(px 0)` -> `(rem 0)` or `zero`. Trying to avoid px where possible 😄

fredrikloberg (approved):
Nice!

src/Nordea/Components/NumberInput.elm
fonnes: Yes, worked great!

src/Nordea/Components/NumberInput.elm
```diff
@@ -141,6 +147,10 @@ getStyles config =
     in
     [ fontSize (rem 1)
     , height (rem 3)
+    , pseudoElement "-webkit-outer-spin-button" [ display none ]
```
fshauge, line 150: https://developer.mozilla.org/en-US/docs/Web/CSS/::-webkit-outer-spin-button

Looks rough when it comes to compatibility. Are there any other solutions?

fonnes: That's why we also have this for firefox :) `property "-moz-appearance" "textfield"`

fonnes: I have at least tested in chrome, safari and mozilla and it hides it fine

PR #106, opened 2022-01-07
Fix styling for dropdown filter

fonnes:
- Add cross to dropdown filter, which deletes search string when clicked
- Add scroll to the list of elements for the search, rather than the search box and the list of results. This must also be fixed in leasing.
- Add cursor pointer to the icon in the search box
- Add padding to input field so it is not possible to write search string over the icon in the search box

Now:
image

With scroll now:
image

Previous:
image

### DIFF
--- a/explorer/src/Config.elm
+++ b/explorer/src/Config.elm
@@ -78,12 +78,7 @@ update msg config =
         OnClickClearSearchComponentInput ->
             let
                 emptySearchString =
-                    case config.searchComponentInput of
-                        Just string ->
-                            Just ""
-
-                        Nothing ->
-                            Nothing
+                    config.searchComponentInput |> Maybe.map (\_ -> "")
             in
             { config | searchComponentInput = emptySearchString }
 
--- a/explorer/src/Config.elm
+++ b/explorer/src/Config.elm
@@ -18,7 +18,7 @@ type alias Config =
     , searchComponentInput : Maybe String
     , searchHasFocus : Bool
     , isFeatureBoxOpen : Bool
-        , isProgressBarCompleted : Bool
+    , isProgressBarCompleted : Bool
     }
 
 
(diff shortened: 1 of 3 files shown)


explorer/src/Config.elm
```diff
@@ -74,6 +75,18 @@ update msg config =
         SearchComponentFocus value ->
             { config | searchHasFocus = value }
 
+        OnCross ->
```
fonnes, line 78: Would love some suggestion for a better name here! :D

src/Nordea/Components/DropdownFilter.elm
```diff
@@ -37,6 +37,7 @@ type alias DropdownFilterProperties a msg =
     { searchItems : List (ItemGroup a)
     , onInput : String -> msg
     , onSelectedValue : Item a -> msg
+    , onCross : msg
```
fredrikloberg, line 40: What about something like this?

```suggestion
    , onClickClearInput : msg
```

explorer/src/Config.elm
fredrikloberg: `OnClickClearSearchComponentInput` or something? 😄 Since this is the global msgs of the UIExplorer, it is nice to also indicate which component it affects

src/Nordea/Components/DropdownFilter.elm
```diff
@@ -194,16 +200,19 @@ inputSearchView hasFocus searchString onInput onFocus =
...
                 ]
           in
-          if hasFocus then
-            Icon.chevronUp attributes
-
+          if String.length searchString > 0 then
+              Icon.cross (attributes ++ [Events.onClick onCross])
           else
-            Icon.chevronDown attributes
+            if hasFocus then
```
fredrikloberg, line 211: missing a format?

fredrikloberg (approved):
Looks great 🤩

explorer/src/Config.elm
```diff
@@ -74,6 +75,18 @@ update msg config =
...
 
+        OnClickClearSearchComponentInput ->
+            let
+                emptySearchString =
+                    case config.searchComponentInput of
+                        Just string ->
+                            Just ""
+
+                        Nothing ->
+                            Nothing
```
fredrikloberg, line 86: Little nitpick:

```suggestion
                    config.searchComponentInput |> Maybe.map (\_ -> "")
```

perok (approved):
Awesome! :100: :raised_hands:

PR #108, opened 2022-01-12
Fix the card padding

NappiLe:
Change padding for the Card component from 2rem to 1.5rem


anisha-deverashetty (approved):
👍

fredrikloberg (commented):
I see that the `cardTitle` sub component assumes that the padding is `2rem`, so that also needs to be updated. See line 67.

Also if you want, the `cardContentContainer` function can be deleted as it is not used.

Other than that, looks good! 🚀

NappiLe:
@fredrikloberg : I fixed it

PR #109, opened 2022-01-13
Add correct customer service email

mathilwa:
Based on the locale we need to show different customer service emails

### DIFF
--- a/src/Nordea/Components/Error.elm
+++ b/src/Nordea/Components/Error.elm
@@ -100,6 +100,9 @@ viewActionForInternalServerError config =
     let
         errorActionText =
             texts.internalServerError.action |> config.translate
+
+        customerServiceEmail =
+            texts.customerServiceEmail |> config.translate
     in
     Text.bodyTextLight
         |> Text.view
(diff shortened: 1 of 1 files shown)


hungqt (approved):
🚀

src/Nordea/Components/Error.elm
```diff
@@ -112,12 +115,12 @@ viewActionForInternalServerError config =
             [ Html.text errorActionText
             , FlatLink.default
                 |> FlatLink.view
-                    [ href "mailto: kundeservice.nfe@nordea.com"
+                    [ href (String.append "mailto" customerServiceEmail)
```
trysnetron, line 118:
```suggestion
                    [ href (String.append "mailto:" customerServiceEmail)
```

vetlesolgaard, line 118: Do we need the `:` in the link? 😁
```suggestion
                    [ href (String.append "mailto:" customerServiceEmail)
```

mathilwa: ah, ofc! it got removed!

PR #115, opened 2022-01-19
Add spinner with custom width

hungqt:
Added an alternative to Spinner component without defined standard width. This way we can use it places like inside buttons etc with custom widths without having to use `Css.important` to overwrite.

image


fredrikloberg (approved):
Great idea! 🤩

PR #121, opened 2022-01-24
[NB: Breaking changes] Allow english text translations

mathilwa:
- In all translations there is added an english text translation
- The Error pages now have to send in their support email by default. This is because it does not make sense to relate email to language (country, yes, language, no), and also, it is very likely that we have different customer support emails for different systems 😊

### DIFF
--- a/src/Nordea/Components/Error.elm
+++ b/src/Nordea/Components/Error.elm
@@ -1,4 +1,4 @@
-module Nordea.Components.Error exposing (internalServerError, pageNotFound, view, withCustomSupportEmail)
+module Nordea.Components.Error exposing (internalServerError, pageNotFound, view)
 
 import Css
     exposing
@@ -23,6 +23,7 @@ import Css
         )
 import Html.Styled as Html exposing (Attribute, Html)
 import Html.Styled.Attributes as Attributes exposing (css, href)
+import Maybe.Extra exposing (isJust)
 import Nordea.Components.FlatLink as FlatLink
 import Nordea.Components.Text as Text
 import Nordea.Html exposing (showIf)
@@ -33,12 +34,12 @@ import Svg.Styled.Attributes as SvgAttrs exposing (d)
 
 
 type alias Translation =
-    { no : String, se : String, dk : String } -> String
+    { no : String, se : String, dk : String, en : String } -> String
 
 
 type alias ErrorConfig =
     { errorType : ErrorType
-    , customSupportEmail : Maybe String
+    , supportEmail : Maybe String
     , translate : Translation
     }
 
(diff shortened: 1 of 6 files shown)


src/Nordea/Components/Error.elm
```diff
@@ -52,38 +53,33 @@ type Error msg
...
 
 
-internalServerError : Translation -> Error msg
-internalServerError translate =
+internalServerError : Translation -> String -> Error msg
+internalServerError translate supportEmail =
     init
         { errorType = InternalServerError
-        , customSupportEmail = Nothing
+        , supportEmail = Just supportEmail
```
fonnes, line 60: Would it be possible to model ErrorConfig to be either a InternalServerErrorConfig or a PageNotFoundErrorConfig, so we don't need to have supportEmail as a Maybe field? It would only be available to InternalServerErrorConfig

fonnes (approved):

src/Nordea/Components/Error.elm
```diff
@@ -131,57 +129,59 @@ view attributes children (Error config) =
...
+                        ]
+                    ]
+                    [ Html.text errorActionText
+                    , FlatLink.default
+                        |> FlatLink.view
+                            [ href ("mailto:" ++ supportEmail)
+                            , css [ display inlineBlock ]
+                            ]
+                            [ Html.text supportEmail ]
                     ]
```
fredrikloberg, line 155:
```suggestion
    config.supportEmail
        |> viewMaybe (\supportEmail ->
            Text.bodyTextLight
                |> Text.view
                    [ css
                        [ maxWidth (rem 30)
                        , paddingRight (rem 0.25)
                        , textAlign center
                        ]
                    ]
                    [ Html.text errorActionText
                    , FlatLink.default
                        |> FlatLink.view
                            [ href ("mailto:" ++ supportEmail)
                            , css [ display inlineBlock ]
                            ]
                            [ Html.text supportEmail ]
                    ]
            )
```

src/Nordea/Components/ProgressBarStepper.elm
```diff
@@ -80,5 +80,6 @@ strings =
         { no = "Neste: "
         , se = "Nästa: "
         , dk = "Næste: "
+        , en = "Next"
```
fredrikloberg, line 83:
```suggestion
        , en = "Next: "
```

src/Nordea/Components/Error.elm
mathilwa: Any suggestions on how to do that in a smooth way?

fredrikloberg: Good idea! But in that case, if the configs get very different, would it make sense to just split them completely? Separate components for internal server error and page not found? They could still live in the same file though, and share sub views if needed.

fredrikloberg (approved):

src/Nordea/Components/Error.elm
fonnes: Something like this maybe? 🤔

`type ErrorConfig
    = InternalServerErrorConfig
        { errorType : ErrorType
        , supportEmail : Maybe String
        , translate : Translation
        }
    | PageNotFoundErrorConfig
        { errorType : ErrorType
        , translate : Translation
        }`

fredrikloberg: With that, the `ErrorType` is redundant and can be removed 😀

mathilwa: I really like the idea of two separate views, since we have this viewAction thing that is not supposed to be shown on PageNotFoundError anyway. I will try to do some changes here 😊 I'll come back to you!

PR #123, opened 2022-01-25
Add withDisabled to TextInput

NappiLe:
Add isDisabled to TextInput


fredrikloberg (commented):
I suggest to not add a property for this, as we can already set it as disabled by sending in `disabled True` as an attribute: 

```elm
TextInput.init "Text"
    |> TextInput.view [ disabled True ]
```

I see that this doesn't currently work when the input has a search icon because of the wrapping div, but that can be fixed the same way Checkbox and Dropdown does it 🙂

PR #129, opened 2022-02-15
Add width to GroupLabel style

hungqt:
Useful when we use the `Label.withRequirednessHint`. With (pct 100) it pushes the hint to the end.

Example:
image

### DIFF
--- a/src/Nordea/Components/Label.elm
+++ b/src/Nordea/Components/Label.elm
@@ -11,7 +11,24 @@ module Nordea.Components.Label exposing
     , withShowFocusOutline
     )
 
-import Css exposing (Style, border, column, displayFlex, flexBasis, flexDirection, flexWrap, marginInlineEnd, marginInlineStart, padding, pct, pseudoClass, rem, width, wrap)
+import Css
+    exposing
+        ( Style
+        , border
+        , column
+        , displayFlex
+        , flexBasis
+        , flexDirection
+        , flexWrap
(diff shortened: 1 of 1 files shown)


src/Nordea/Components/Label.elm
```diff
@@ -137,7 +138,7 @@ view attrs children (Label config) =
                 ]
                 attrs
                 ((legend
-                    [ css [ padding (rem 0), flexBasis (pct 100), Css.Global.children [ everything [ flexBasis (pct 100) ] ] ] ]
+                    [ css [ width (pct 100), padding (rem 0), flexBasis (pct 100), Css.Global.children [ everything [ flexBasis (pct 100) ] ] ] ]
```
fredrikloberg, line 141: I thought the flexBasis would do the same, but maybe it is shrinking? 🤔
```suggestion
                    [ css [ flexShrink (num 0), padding (rem 0), flexBasis (pct 100), Css.Global.children [ everything [ flexBasis (pct 100) ] ] ] ]
```

fredrikloberg (approved):

src/Nordea/Components/Label.elm
hungqt: `flexShrink` doesnt seem to work

fredrikloberg: all right, but does width work if we remove flexBasis? If so, I think we should remove it

Issue #144, opened 2022-03-24
Add "small" version of input fields


TheaSommersletta:
All types of input fields should have a small and large version

Large = 48px height
Small = 41 px height

morteako:
@TheaSommersletta Merged (perhaps a bit early 😅). But this can be closed if you are happy with the result

PR #150, opened 2022-03-28
#144 - Small Textinput

morteako:
Add "small" version of input fields #144

image

image

The difference between 48px(3 rem) and 41px(2.5625 rem) is not that, I guess.


fredrikloberg (approved):
Looks good!

fonnes (approved):
Nice! 🤩

PR #151, opened 2022-03-29
Refactor ideas

morteako:
Some low hanging fruit. Some more controversial.
Open for discussion

### DIFF
--- a/src/Nordea/Components/FileUpload.elm
+++ b/src/Nordea/Components/FileUpload.elm
@@ -57,12 +57,12 @@ import Html.Styled as Html exposing (Attribute, Html)
 import Html.Styled.Attributes exposing (accept, css, multiple, type_, value)
 import Html.Styled.Events as Events
 import Json.Decode as Decode
+import Nordea.Components.Common exposing (Translation)
 import Nordea.Components.Text as Text
 import Nordea.Html exposing (showIf, styleIf, viewMaybe)
 import Nordea.Resources.Colors as Colors
 import Nordea.Resources.Icons as Icon
 import Nordea.Themes as Themes
-import Nordea.Types exposing (Translation)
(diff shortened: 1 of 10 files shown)


src/Nordea/Components/FileUpload.elm
```diff
@@ -396,6 +395,15 @@ onSelectFiles msg =
...
 
+strings :
+    { uploadDescription1 : Translation
+    , uploadDescription2 : Translation
+    , uploadDescription3 : Translation
+    , acceptedFileTypes : Translation
+    , and : Translation
+    , dropToUploadFile : Translation
+    , uploadedFiles : Translation
+    }
```
fredrikloberg, line 406: I'm a little skeptical about the value of adding type annotations to the translations record. It's not too bad here in the component library I guess, but in PartnerHub we have many more translations per file. My concern is that it adds mostly noise and not much value, since it's a hardcoded structure and Elm has "perfect" type inference.

What are your thoughts about it?

fredrikloberg (approved):
Nice looks good! just one comment we can discuss

src/Nordea/Components/FileUpload.elm
morteako: I kinda agree. I think I just did it to get one less warning.
Should I just remove this annotation, or remove the whole Translation type?

fredrikloberg: I think we can keep the translation type, and just remove the annotation. Is the warning from elm-review? Maybe we can add an exception for the translations record?

morteako: Seems like just some stuff in vs-code that complained. We can just ignore that.
Elm-review does not complain.

I removed it now 👍

PR #152, opened 2022-04-01
Add attributes to check icon

Eivinddh:
I noticed that the ✔️ icon wasn't wrapped  `iconContainer` and didn't take attributes as parameters.

I'll also look into stories for icons


vetlesolgaard (approved):
Nice! 🤩

PR #153, opened 2022-04-11
Remove spesific css from cross icon

Eivinddh:
I should have updated the usages as well, at least according to https://github.com/search?l=Elm&q=org%3ASGFinansAS+%22cross+%22&type=Code 

Doesn't seem like any other application uses it?


hungqt (approved):
LGTM Nice 🤩

PR #164, opened 2022-04-28
PAR-Add-card-infobox: Addded infoBox component to Card component

vetlesolgaard:
- Adds an `infoBox` component to the `Card` component concept

Screenshot 2022-04-28 at 11 00 04


fredrikloberg (approved):
Great!

PR #201, opened 2022-05-23
Add outer and inline link variants of flatlink

### LINKED ISSUE #189: New: Outer link
Standalone links are used on their own directly following content. They should not be used within sentences or paragraphs. Standalone links are the default link style and only have an underline in the hover state.

The standalone link component can be paired with an icon. Use 16px icons and place them to the right of the link. Icons should always be the same color as the link text.

Image

trysnetron:
Closes #189 #188 

bilde
bilde


fredrikloberg (commented):
Just a thought: instead of creating a new component, what do you think of extending the existing link component instead? Changing the display from `flex` to `inline-flex` will make it inline.

src/Nordea/Components/InlineLink.elm
```diff
@@ -0,0 +1,105 @@
...
+-- CONFIG
+
+
+type alias Config =
+    { styles : List Style, isDisabled : Bool }
+
+
+type InlineLink
+    = InlineLink Config
+
```
vetlesolgaard, line 45: Do we need the `Config` record? 😊
```suggestion
type InlineLink
    = InlineLink { isDisabled : Bool }
```

src/Nordea/Components/FlatLink.elm
```diff
@@ -103,6 +109,47 @@ view attributes children (FlatLink config) =
...
+                , Themes.color Themes.PrimaryColor Colors.blueNordea
+                , textDecoration none
+                , hover
+                    [ Themes.color Themes.PrimaryColorLight Colors.blueDeep
+                    , textDecoration underline
+                    ]
+                , focus [ Themes.color Themes.PrimaryColorLight Colors.blueDeep ]
+                , visited [ Themes.color Themes.PrimaryColorLight Colors.purple ]
+                ]
+            , isDisabled = Just False
```
morteako, line 147: kinda out of scope, but should this Maybe Bool just be Bool with default False instead of Nothing?

src/Nordea/Components/FlatLink.elm
```diff
@@ -103,6 +109,47 @@ view attributes children (FlatLink config) =
         children
 
 
+inline : String -> String -> Html msg
+inline url text =
```
fredrikloberg, line 113: I suggest an interface like this. Makes it possible to add aria-tags, onclick tracking etc. on demand

```suggestion
inline : List (Attribute msg) -> List (Html msg) -> Html msg
inline attrs children =
```

src/Nordea/Components/FlatLink.elm
```diff
@@ -103,6 +109,47 @@ view attributes children (FlatLink config) =
...
+        (FlatLink
+            { styles =
+                [ Css.display inlineFlex
+                , Css.fontWeight Css.inherit
+                , Themes.color Themes.PrimaryColor Colors.blueNordea
+                , hover [ Themes.color Themes.PrimaryColorLight Colors.blueDeep ]
+                , focus [ Themes.color Themes.PrimaryColorLight Colors.blueDeep ]
+                , visited [ Themes.color Themes.PrimaryColorLight Colors.purple ]
+                ]
+            , isDisabled = Just False
```
fredrikloberg, line 125: With the interface change suggested above, this can be controlled by:
```suggestion
            , isDisabled = Just (List.member (disabled True) attrs)
```

src/Nordea/Components/FlatLink.elm
```diff
@@ -103,6 +109,47 @@ view attributes children (FlatLink config) =
...
+inline url text =
+    view [ href url ]
+        [ Html.text text ]
+        (FlatLink
+            { styles =
+                [ Css.display inlineFlex
+                , Css.fontWeight Css.inherit
+                , Themes.color Themes.PrimaryColor Colors.blueNordea
+                , hover [ Themes.color Themes.PrimaryColorLight Colors.blueDeep ]
+                , focus [ Themes.color Themes.PrimaryColorLight Colors.blueDeep ]
```
fredrikloberg, line 122: I think the theme colors here are mixed, or is it intentional? (see https://github.com/SGFinansAS/elm-components#themes)

src/Nordea/Components/FlatLink.elm
```diff
@@ -103,6 +109,47 @@ view attributes children (FlatLink config) =
...
+                , visited [ Themes.color Themes.PrimaryColorLight Colors.purple ]
+                ]
+            , isDisabled = Just False
+            , variant = Default
+            }
+        )
+
+
+outer : String -> String -> Html msg
+outer url text =
```
fredrikloberg, line 132: Suggest the same interface here

src/Nordea/Components/FlatLink.elm
trysnetron: lgtm

src/Nordea/Components/FlatLink.elm
trysnetron: Matches the design sketches in the [issue](https://github.com/SGFinansAS/elm-components/issues/188) 🤔

fredrikloberg: I think it always uses the fallback colors here because no theme variables are defined by default in the component library.

I think this is more correct:
```suggestion
                , Themes.color Themes.PrimaryColorLight Colors.blueNordea
                , hover [ Themes.color Themes.PrimaryColor Colors.blueDeep ]
                , focus [ Themes.color Themes.PrimaryColor Colors.blueDeep ]
```

fredrikloberg (approved):
Just the theme variables, else looks good!

PR #208, opened 2022-05-23
radio button small

MistyBui:


src/Nordea/Components/RadioButton.elm
```diff
@@ -65,6 +66,11 @@ import Nordea.Resources.Colors as Colors
 import Nordea.Themes as Themes
 
 
+type Size
+    = DefaultSize
+    | Small
```
fredrikloberg, line 71: We have used these terms for this in the other components, should we stick with that?
```suggestion
type Variant
    = Standard
    | Small
```

src/Nordea/Components/RadioButton.elm
```diff
@@ -96,9 +103,15 @@ init name label onCheck =
         , isSelected = False
         , appearance = Standard
         , showError = False
+        , size = DefaultSize
         }
 
 
+small : RadioButton msg -> RadioButton msg
+small (RadioButton config) =
```
fredrikloberg, line 111:
```suggestion
withSmallSize : RadioButton msg -> RadioButton msg
withSmallSize (RadioButton config) =
```

fredrikloberg (approved):
Looks good!

hhaga (approved):
Looks good 😎 
Minor: I see `Standard` has been used instead of `DefaultSize` elsewhere. Could consider changing to that to make it consistent 😊

Claudio-Pollice:
> 

@hhaga How do we deal with 2 different types both having a "Standard" value? That is the case here.
In general, should we avoid terms that can be mistaken for different types was our thinking?

src/Nordea/Components/RadioButton.elm
Claudio-Pollice: Related to @hhaga 's comment. When we tried following the convention we got conflicting values "Standard" of more than 1 type. We didn't see how to resolve that, other than having different values. And then we thought "should we in general go for expressive / specific names rather than broad / general names that can have overlap and conflict" ?

This might just be our lack of knowledge regarding Elm here :D

fredrikloberg: Ahh, I see! So, instead of adding a new type and field for size, what about adding a `Small` variant to the existing Appearance type? Do you think that would make sense?

Claudio-Pollice: @hhaga @fredrikloberg tagging both since same discussion

Specifically our issue is:

`appearance: Standard
, variant: Standard
`

We can't solve this with doing 


`appearance: Appearance.Standard
, variant: Variant.Standard
`

fredrikloberg: I suggest to extend the exisitng Appearance type like so:
```elm
type Appearance
    = Standard
    | Simple
    | ListStyle
    | Small
```

Claudio-Pollice: @fredrikloberg adding it to Appearance makes sense if we never want to combine listStyle + small.
That coupling wasn't something we thought was in the requirements, but it would fix the current issue.
Do you think we should pay that limitation as the price?

fredrikloberg: Yes, I think that makes sense. The ListStyle is a very different styling anyway

PR #209, opened 2022-05-23
191 - Informationdetails

morteako:
Part of #191

### DIFF
--- a/src/Nordea/Components/InformationDetails.elm
+++ b/src/Nordea/Components/InformationDetails.elm
@@ -7,8 +7,8 @@ module Nordea.Components.InformationDetails exposing
     , value
     )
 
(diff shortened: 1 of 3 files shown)


src/Nordea/Components/InformationDetails.elm
```diff
@@ -0,0 +1,103 @@
...
+        (css
+            [ displayFlex
+            , flexWrap wrap
+            , children
+                [ everything
+                    [ pseudoClass "not(:last-child)" [ marginRight (rem 0.5) ] ]
+                ]
+            ]
+            :: attrs
+        )
```
fredrikloberg, line 85: I think we can delete this
```suggestion
```

trysnetron (approved):
TLDRLGTM 😎 👍

PR #210, opened 2022-05-23
Range Slider Component

hungqt:
image

image

image


explorer/src/Config.elm
```diff
@@ -24,6 +24,7 @@ type alias Config =
     , isHoveringFileUpload : Bool
     , selectedFiles : List File
     , infoLabel : InfoLabel
+    , sliderInputValue : String
```
morteako, line 27: could this be float? Would that simplify/remove the need of the number parsing below?

fredrikloberg (commented):
Is it possible to split out the slider from the number input and label? I.e., this component only consists of the slider, and the label is added in the same manner as it is for the other input components

morteako:
image

src/Nordea/Components/Slider.elm
```diff
@@ -0,0 +1,139 @@
+module Nordea.Components.Slider exposing (..)
+
+-- CONFIG
```
morteako, line 3: delete or move? 😅 probably delete 🙃🚀

src/Nordea/Components/Slider.elm
```diff
@@ -0,0 +1,139 @@
...
+            ((((config.value |> String.toFloat |> Maybe.withDefault 0) - config.min) * 100) / (config.max - config.min)) |> String.fromFloat
+
+        cloudBlue =
+            "#DCEDFF"
+
+        nordeaBlue =
+            "#0000A0"
+
+        gradientValue =
+            "linear-gradient(to right," ++ nordeaBlue ++ " 0% " ++ visibleWidth ++ "%, " ++ cloudBlue ++ " " ++ visibleWidth ++ "%" ++ " 100% )"
```
fredrikloberg, line 135: Can we use the colors from Colors.elm here instead of hardcoded strings? We used to have a `colorToString` function, so that may be used to convert them to strings if needed

morteako:
The component looks great 🍾🍾🏆🍺

explorer/src/Config.elm
hannabergljot: We did it like this because the Html input type requires a string, but we could also parse it the other way I guess:) Any opinions on what way of parsing is preferred?

src/Nordea/Components/Slider.elm
hannabergljot: Nice, we'll use that!😁

fredrikloberg: And also, support for themes would be nice 😀

explorer/src/Config.elm
morteako: My preference is float to string, because that is safer, less error handling and I feel like it actually is a number value that we use, so it feels cleaner to me to have a number and when you display it then you convert it to a string.
Vs having a string that you hope is a number

hungqt: Agreed :D we'll change it to Float

hungqt:
Split Range.elm as its own component out of Slider.elm now. Now Range works as only the range-input, while Slider is the range-input together with the label and optional NumberInput.

src/Nordea/Components/Range.elm
```diff
@@ -0,0 +1,153 @@
...
+import Nordea.Resources.Colors as Colors
+import Nordea.Themes as Themes
+
+
+type alias Config msg =
+    { value : Float
+    , min : Float
+    , max : Float
+    , step : Maybe Float
+    , onInput : String -> msg
```
fredrikloberg, line 46: As the onInput event returns a string and the `string -> float` conversion may fail, I think we need a decoder to achieve this. Check out the decoder and event in Dropdown.elm

```suggestion
    , onInput : Float -> msg
```

src/Nordea/Components/Range.elm
```diff
@@ -0,0 +1,153 @@
...
+view : List (Attribute msg) -> Range msg -> Html msg
+view attributes (Range config) =
+    div attributes
+        [ input
+            [ name "rangeInput"
+            , type_ "range"
+            , Html.Styled.Attributes.value (config.value |> String.fromFloat)
+            , Html.Styled.Attributes.min (config.min |> String.fromFloat)
+            , Html.Styled.Attributes.max (config.max |> String.fromFloat)
+            , Html.Styled.Attributes.step (config.step |> Maybe.map String.fromFloat |> Maybe.withDefault "1")
```
fredrikloberg, line 85: Or prefixed `Attrs.` if the names are taken
```suggestion
            , value (config.value |> String.fromFloat)
            , min (config.min |> String.fromFloat)
            , max (config.max |> String.fromFloat)
            , step (config.step |> Maybe.map String.fromFloat |> Maybe.withDefault "1")
```

src/Nordea/Components/Range.elm
```diff
@@ -0,0 +1,153 @@
...
+            , config.onInput |> onInput
+            , css [ sliderStyle (Range config) ]
+            ]
+            []
+        , showIf config.showInterval
+            (div [ css [ displayFlex, flexDirection row, justifyContent spaceBetween, margin2 (rem 1) (rem 0.25) ] ]
+                (List.range (config.min |> ceiling) (config.max |> ceiling) |> List.map (\number -> span [] [ Html.text (String.fromInt number) ]))
+            )
```
fredrikloberg, line 93: Not 100% sure about the change to ol/li being semantically correct here. What do you think?
```suggestion
        , ol 
            [ css 
                [ displayFlex
                , flexDirection row
                , justifyContent spaceBetween
                , margin2 (rem 1) (rem 0.25) 
                ] 
            ]
            (List.range (ceiling config.min) (ceiling config.max) 
                |> List.map (\number -> li [] [ Html.text (String.fromInt number) ])
            )
            |> showIf config.showInterval
```

src/Nordea/Components/Range.elm
```diff
@@ -0,0 +1,153 @@
...
+            [ width (pct 100)
+            , height (rem 0.25)
+            , property "background-color" "transparent"
+            ]
+        , pseudoElement "-webkit-slider-thumb"
+            [ property "-webkit-appearance" "none"
+            , width (rem 1.5)
+            , height (rem 1.5)
+            , borderRadius (pct 100)
+            , Themes.backgroundColor Themes.SecondaryColor Colors.blueDeep
```
fredrikloberg, line 115:
```suggestion
            , Themes.backgroundColor Themes.PrimaryColor Colors.blueDeep
```

src/Nordea/Components/Range.elm
```diff
@@ -0,0 +1,153 @@
...
+        , pseudoElement "-moz-range-track"
+            [ width (pct 100)
+            , height (rem 0.25)
+            , backgroundColor transparent
+            ]
+        , pseudoElement "-moz-range-thumb"
+            [ width (rem 1.25)
+            , height (rem 1.25)
+            , borderRadius (pct 100)
+            , Themes.backgroundColor Themes.SecondaryColor Colors.blueDeep
```
fredrikloberg, line 129:
```suggestion
            , Themes.backgroundColor Themes.PrimaryColor Colors.blueDeep
```

src/Nordea/Components/Range.elm
```diff
@@ -0,0 +1,153 @@
...
+
+        nordeaBlue =
+            Themes.colorVariable Themes.PrimaryColorLight Colors.blueNordea
+
+        gradientValue =
+            "linear-gradient(to right," ++ nordeaBlue ++ " 0% " ++ visibleWidth ++ "%, " ++ cloudBlue ++ " " ++ visibleWidth ++ "%" ++ " 100% )"
+    in
+    Css.batch
+        [ property "background" gradientValue
+        ]
```
fredrikloberg, line 153:
```suggestion
    property "background" gradientValue
```

src/Nordea/Components/Slider.elm
```diff
@@ -0,0 +1,121 @@
...
+    Slider { config | showNumberInput = value }
+
+
+
+-- VIEW
+
+
+view : List (Attribute msg) -> Slider msg -> Html msg
+view attributes (Slider config) =
+    div ([] ++ attributes)
```
fredrikloberg, line 100:
```suggestion
    div attributes
```

src/Nordea/Components/Range.elm
```diff
@@ -0,0 +1,153 @@
...
+
+
+-- VIEW
+
+
+view : List (Attribute msg) -> Range msg -> Html msg
+view attributes (Range config) =
+    div attributes
+        [ input
+            [ name "rangeInput"
```
fredrikloberg, line 80: Do we need the name? I see it is used with the label's `for` attribute in Slider, but I think the `for` attribute should be used with id's instead of names. And id's should generally be avoided in a component because it means that the component can only be used once per page, which may not be expected. I suggest to wrap the input in the label instead


```suggestion
            [
```

fredrikloberg (approved):
Noice!

src/Nordea/Components/Range.elm
hungqt: I think its atleast more semantically correct than the div and spans right now either way 😅 Good suggestion!

src/Nordea/Components/Range.elm
hungqt: I want to remove these `name` and `for` attributes aswell 🤔 but how can i go on about it when the use in Slider is structured like this?

```
div[][
    div[][
        label[][]
        , numberInput[][]
    ]
    , Rangeinput [][]
]
```

hungqt: Cant find a way to wrap the label around the rangeInput with this setup

fredrikloberg: This is the most semantically correct setup I can think of for this. I think the label for the two inputs may be visually hidden, but needs to be there for accessibility reasons

```elm
fieldset []
    [ legend [] [ Html.text "The label for the whole component" ]
    , label [] [ Html.text "Label for the numberinput", numberInput [] [] ]
    , label [] [ Html.text "Label for the rangeinput", rangeInput [][] ]
]
```

PR #215, opened 2022-05-27
Add tooptip disappear timing

NappiLe:
A tooltip should disappear after sometime

### DIFF
--- a/explorer/src/Stories/Tooltip.elm
+++ b/explorer/src/Stories/Tooltip.elm
@@ -68,4 +68,28 @@ stories =
                     ]
           , {}
           )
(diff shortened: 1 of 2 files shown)


morteako:
I think it would be nice to merge the to fields 
    , isDisappear : Bool
    , visibleDurationMs : Maybe Float
into one

a custom type like
`type Visibility = Always | DurationMs Float`
and a field      visibility: Visibility
can make it easier to use and remove the possibility of weird states

(The names are just examples and can ofc be changed)

morteako (approved):
Looks great to me!

PR #218, opened 2022-06-14
Disable NumberInput based on config

anisha-deverashetty:


fredrikloberg (commented):
I think we actually landed on not using a `withIsDisabled` function when the standard attribute is sufficient. We do this for the other components as well (TextInput, TextArea, RadioButton, Checkmark, etc.). So I think the bug was that the function should have been removed from the NumberInput component and we should switch to using the attribute in PartnerHub

anisha-deverashetty:
Ok then I will close this PR

PR #224, opened 2022-06-27
Fix: Slider - stupid explorer default styling

hungqt:
Fixes the Slider's default fieldset border styling because Storybook Explorer hides it


hannabergljot (approved):
🙌🙌

PR #226, opened 2022-07-01
Small button variants

Eivinddh:
Made a small button variants that's 32 px = 2 rem high.

**Primary**
Screen Shot 2022-07-01 at 08 35 40

**PrimarySmall**
Screen Shot 2022-07-01 at 08 35 45

**Secondary**
Screen Shot 2022-07-01 at 08 35 50

**SecondarySmall**
Screen Shot 2022-07-01 at 08 35 56

**Tertiary**
Screen Shot 2022-07-01 at 08 36 00

**TertiarySmall**
Screen Shot 2022-07-01 at 08 36 06


fredrikloberg (commented):
The height of the normal buttons is controlled by the padding + content. When setting the height for the small variants without changing the padding, content is slightly "cut" because there is not enough space left.

Screenshot 2022-07-01 at 10 04 35

This is just a nitpicky detail, but I think it may be good to change the normal buttons too and set a specific height (2.5 rem) and remove the vertical padding. I see that the height of the normal buttons is slightly wrong too compared to Figma. It is 40px (2.5rem) in Figma, but is 38,4px in Elm

fredrikloberg (approved):
Just a small comment, else 💯

src/Nordea/Components/Button.elm
```diff
@@ -85,8 +89,11 @@ import Nordea.Themes as Themes
 
 type Variant
     = Primary
+    | PrimarySmall
     | Secondary
+    | SecondarySmall
     | Tertiary
+    | TertiarySmall
     | Card
     | SmallCard
```
fredrikloberg, line 98: is it worth renaming this too for consistency?

fredrikloberg (approved):
Looks good!

src/Nordea/Components/Button.elm
Eivinddh: Yeah, I can do it in a seperate PR ☺️

PR #228, opened 2022-07-05
PAR-Add-hide-if-function

vetlesolgaard:
Adds a `hideIf` function, which works in the opposite way as the existing `showIf` function. Sometimes it is easier to read code when it is phrased as `hideIf` as contrary to `showIf`.


NappiLe:
LGTM 💯

NappiLe (approved):

PR #233, opened 2022-08-01
PAR: Add action button

NappiLe:
circular

### DIFF
--- a/src/Nordea/Components/Button.elm
+++ b/src/Nordea/Components/Button.elm
@@ -1,6 +1,7 @@
 module Nordea.Components.Button exposing
     ( Button
     , Variant(..)
+    , actionButton
     , buttonStyleForExport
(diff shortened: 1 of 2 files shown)


src/Nordea/Components/Button.elm
```diff
@@ -90,6 +95,7 @@ type Variant
     | Card
     | SmallCard
     | FlatLinkStyle
+    | ActionButton
```
fredrikloberg, line 98: Is this a better name?

```suggestion
    | Circular
```

fredrikloberg (approved):
Just a minor naming suggestion, else 💯

src/Nordea/Components/Button.elm
NappiLe: Yea, I will rename it

PR #240, opened 2022-08-23
Rename small card for consistent naming

Eivinddh:
I renamed the constructor function as well which will not provide backwards compatibilty 🤔 Should it remain inconsistent?


fredrikloberg (approved):
Looks good!

PR #241, opened 2022-08-23
Par 1821 small fixes to dropdown

vetlesolgaard:
- [x] Design review

**Changes**
- There is too much space in the drop-down below the search result
- Make the “X” smaller (1rem)
- Looks like its the wrong font once you’ve selected something

❔**Questions**
- Seems a bit weird that the `Tooltip` component is used for showing search content for the `DropdownFilter`🤔 Is this something we should change?

📸**Screenshot:**
Screenshot 2022-08-23 at 16 30 30

### DIFF
--- a/src/Nordea/Components/DropdownFilter.elm
+++ b/src/Nordea/Components/DropdownFilter.elm
@@ -197,7 +197,7 @@ view attrs (DropdownFilter config) =
                         , top (pct 50)
                         , right (rem 0.75)
                         , transforms [ translateY (pct -50) ]
-                        , width (rem 1.125)
+                        , width (rem 1)
                         , cursor pointer
                         ]
                     ]
--- a/src/Nordea/Components/DropdownFilter.elm
+++ b/src/Nordea/Components/DropdownFilter.elm
@@ -29,13 +29,9 @@ import Css
         , deg
         , displayFlex
         , flexDirection
-        , fontSize
-        , fontWeight
         , height
         , hover
-        , int
         , justifyContent
-        , lineHeight
         , listStyle
         , margin2
         , maxHeight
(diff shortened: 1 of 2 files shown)


src/Nordea/Components/DropdownFilter.elm
```diff
@@ -117,7 +116,7 @@ init { onInput, input, onSelect, items } =
         , onSelect = onSelect
         , input = input
         , onFocus = Nothing
-        , hasFocus = True
+        , hasFocus = False
```
fredrikloberg, line 119: Should double check that this doesn't break anything in FF. I think the partnerhub super user login page in FF may rely on this

vetlesolgaard: Nice spot! It seemed not necessary to have this set to `True`. I'll check it out

src/Nordea/Components/DropdownFilter.elm
```diff
@@ -134,17 +133,43 @@ view attrs (DropdownFilter config) =
...
+            let
+                viewHeader text =
+                    Html.li
+                        [ css
+                            [ color Colors.gray
+                            , margin2 (rem 0) (rem 0.625)
+                            , fontSize (rem 0.75)
+                            , fontWeight (int 400)
+                            , lineHeight (rem 1)
+                            ]
```
fredrikloberg, line 145: Do this need to be a custom styled text, or can we use of the the text components? (and maybe add a few overrides). May be worth discussing the header style with Thea, Andreas has already complained that they are somewhat hard to read with the current style

fredrikloberg (approved):
Looks good! Regarding the use of Tooltip, I find that it serves this purpose very well. The only thing is maybe the name of it. The news dropdown in PartnerHub also uses the tooltip for its content

src/Nordea/Components/DropdownFilter.elm
vetlesolgaard: It seems to only be in use in the configuration page, and for the PartnerHub page, both seem to work fine with it being `False` 😊

src/Nordea/Components/DropdownFilter.elm
vetlesolgaard: Used the `TextTinyLight` variant for now, seems like the closest to what we want😊 I'll have a talk with Thea a day she is at the office

PR #243, opened 2022-08-24
Update: set small as standard for all inputs #170

morteako:


src/Nordea/Components/Checkbox.elm
```diff
@@ -81,7 +81,7 @@ type Checkbox msg
 
 
 type Appearance
-    = Standard
+    = Large
     | Simple
     | ListStyle
     | Small
```
fredrikloberg, line 87: Since the long term plan is to delete the large ones completely, is it better to keep the `Standard` appearance (but make it smaller)?

The appearance variants here have very different styling, but the Small/Large are really just different heights of the same style (the standard styling)

fredrikloberg: Maybe you can just delete the Large variants altogether in this PR and then we are done with it? 😀

morteako:
Old. Probably not needed

PR #247, opened 2022-10-11
PAR-1937: Bugfix: Check mimetype for drag n drop files

fredrikloberg:
Files that were drag n dropped were not validated for the mime type.


vetlesolgaard (approved):
🙌

PR #249, opened 2022-10-25
PAR-1878: Added new DropdownOptions component

vetlesolgaard:
**Main task:**
- Created a `DropdownOptions.elm` component to deal with select fields which have checkboxes inside the dropdown. This is a complex component, but we need it for a feature in PartnerHub.

**Additional stuff:**
- I feel that the need to wrap all components in the `Label.elm` component makes this component library a bit harder to use. I think a better solution would be if all components could just be used independently of this.
For the components we update to not having to use the `Label` component the difference would in practice look like this👇
```elm
Label.init (strings.phoneNumber |> translate lang) Label.InputLabel
    |> Label.view
        []
        [ form
            |> SupplierUserForm.phoneNumber
            |> TextInput.init
            |> TextInput.withSmallSize
            |> TextInput.withOnInput (\newValue -> UpdateForm (SupplierUserForm.updatePhoneNumber newValue))
            |> TextInput.view []
        ]
```
vs
```elm
form
    |> SupplierUserForm.phoneNumber
    |> TextInput.init
    |> TextInput.withLabel (strings.phoneNumber |> translate lang)
    |> TextInput.withSmallSize
    |> TextInput.withOnInput (\newValue -> UpdateForm (SupplierUserForm.updatePhoneNumber newValue))
    |> TextInput.view []
```
To mitigate this difficulty of having to use the `Label.elm` component everywhere I refactored the code a little bit👇
- Created a `Hint.elm` component which is supposed to handle everything that goes on below the input fields.
- Created a `RequirednessHint.elm` component which should handle everything related to the requiredness hint in the top right corner of input fields.

**Screenshot**:
https://user-images.githubusercontent.com/11747383/197774903-05641e43-3280-4d10-823e-c31b0df16a95.mov


src/Nordea/Components/DropdownOptions.elm
```diff
@@ -0,0 +1,252 @@
...
+        [ Html.row [ css [ displayFlex, justifyContent spaceBetween, marginBottom (rem 0.2) ] ]
+            [ Common.viewLabel { label = dropdown.label, isError = False } []
+            , dropdown.requirednessHint |> Html.viewMaybe RequirednessHint.view
+            ]
+        , Html.div
+            [ class "nordea-select"
+            , Events.on "focusout" (Decode.succeed (dropdown.onFocus False))
+            , Events.on "focusin" (Decode.succeed (dropdown.onFocus True))
+            , css [ position relative ]
+            , tabindex -1
```
vetlesolgaard, line 190: Had to add `tabindex -1` to be able to receive focus🤔 Is there another way to do it?

fredrikloberg:
Regarding the Label, some drawbacks with the proposed solution as I see it:
* There is a disconnect between how we write the view vs how it is represented in HTML
  * In HTML, label is always wrapped around other inputs 
  * I think this disconnect is a major driver for bad accessibility (UU)
* We essentially have to reimplement the label component for every component it is relevant for
  * Yes we can reuse the styles and bare views, but the config needs to be reimplemented across all components (all required and optional parameters)
  * A change to label may require updating all components that use it
*  We have to predict which components should support label
* May be hard to get a consistent interface
  * A label may wrap one or multiple components (e.g., radio buttons and checkboxes, but it is also used to wrap an input and a slider in the open leasing calculator)
* Label is not only used for inputs, but also other components (e.g., for informational text)
* Components with nested html elements are hard to design
  * What if label needs some css, and the actual input needs some other css? This is trivial if they are separate composable components. This same goes for accessibility props (WAI-ARIA). 


Helper functions is an alternative solution that avoids many of these drawbacks but are still easier to use, for example:
```elm
form
    |> SupplierUserForm.phoneNumber
    |> TextInput.init
    |> TextInput.withSmallSize
    |> TextInput.withOnInput (\newValue -> UpdateForm (SupplierUserForm.updatePhoneNumber newValue))
    |> Label.wrapInput (strings.phoneNumber |> translate lang) []
```

I know NAV's design system implements a hybrid of how we currently have it and your proposal. For the simple cases, their input components support a label prop (just a string), and for all other cases where they require hint etc, they use it as a separate component like we currently do. 


What are you thoughts on this?

src/Nordea/Components/DropdownOptions.elm
fredrikloberg: I think this is the correct way to add focus support to divs (and other elements) according to the spec

fredrikloberg:
Regarding the name of the component, what do you think about `MultiSelectDropdown` or something similar? To me, _options_ doesn't necessarily mean multiple, as options are also used for single select dropdowns

vetlesolgaard:
> Regarding the name of the component, what do you think about `MultiSelectDropdown` or something similar? To me, _options_ doesn't necessarily mean multiple, as options are also used for single select dropdowns

Much better! 🙌

vetlesolgaard:
> Regarding the Label, some drawbacks with the proposed solution as I see it:
> 
> * There is a disconnect between how we write the view vs how it is represented in HTML
>   
>   * In HTML, label is always wrapped around other inputs
>   * I think this disconnect is a major driver for bad accessibility (UU)

My impression is that it is not a strict rule accessibility wise that labels need to contain inputs. Just linking the first relevant search hit: https://www.w3.org/WAI/tutorials/forms/labels/#associating-labels-explicitly

> We have to predict which components should support label

I don't think this will be an issue as it is pretty set in figma which inputs have labels. It will also be very easy to add afterwards, if there is a need to retrofit.

> May be hard to get a consistent interface

For groups of inputs, like radio groups, we'll have to make it part of the label component or in its own component which is then used in the relevant components. I don't think this will be a problem and might actually be more semantically correct as a grouplabel (`legend`) and a label for an input is not the same html element.

> Helper functions is an alternative solution that avoids many of these drawbacks but are still easier to use, for example:

Good idea, but I think labels are a so integral part of our components that they should be a part of them🙏

An advantage to baking them in with the component is that it will be a much better interface for us as programmers. 
If we look to big established component libraries as [NAV](https://aksel.nav.no/designsystem/komponenter/textarea) and [Entur](https://design.entur.org/komponenter/skjemaelementer/textfield), I see that they also have label as part of their components and it does not exist independently.

vetlesolgaard:
This could be a good theme for town hall, as it will affect everyone working with the component library? 😊

src/Nordea/Components/Common.elm
```diff
@@ -54,109 +34,26 @@ type alias CharCounter =
...
-        toI18NString requirednessHint =
-            case requirednessHint of
-                Mandatory translate ->
-                    translate strings.mandatory
+viewLabel : LabelProperties -> List (Attribute msg) -> Html msg
+viewLabel { label, isError } attrs =
+    Text.textSmallLight
+        |> Text.view
+            ([ css [ color Colors.redDark |> styleIf isError ] ] ++ attrs)
+            [ Html.label [] [ Html.text label ] ]
```
fredrikloberg, line 48: This label is not semantically connected to any form controls. It should either have a `for` prop referencing the id of the element, or be wrapped around it. For reusable components, I prefer the latter as you avoid having to enforce that the id is unique across the page. 

I would also suggest to have the Text inside the label rather that having the label inside the text. When using the viewLabel function, one should expect the attributes to be set on the html label and not the span.

vetlesolgaard: I'll add the `for` property, good catch🙏  To enforce the uniqueness of it we could prefix with type of field and postfix with random id like Entur has done.
Was a bit uncertain on where the label should be. Sound like a good idea to put it outside the `Text`.

fredrikloberg: Random is cumbersome in Elm tho, all inputs suddenly require an update roundtrip and state 😛

vetlesolgaard: 😬 I'll try to figure out something that works 🙏

src/Nordea/Components/DropdownOptions.elm
```diff
@@ -0,0 +1,252 @@
...
+            ]
+            [ Html.select
+                [ css [ display none ] ]
+                (dropdown.options |> List.map (\option -> Html.option [ name option.name ] [ Html.text option.label ]))
```
fredrikloberg, line 185: I wonder if a native accordion with a fieldset may better represent what we are doing here.

Check out this example https://jsfiddle.net/hvto259e/

This is natively supported, thus semantically correct, and does what we want out of the box. We only have to style it to look how we want

```html
<details>
  <summary>This is the collapsed label</summary>

  <fieldset>
    <label> 
        some label
        <input type="checkbox" />
    </label>
        <label> 
        some other label
        <input type="checkbox" />
    </label>
  </fieldset> 
</details>
```

src/Nordea/Components/DropdownOptions.elm
```diff
@@ -0,0 +1,252 @@
...
+                    , displayFlex
+                    , alignItems center
+                    , justifyContent spaceBetween
+                    , focus
+                        [ Css.property "box-shadow" ("0rem 0rem 0rem 0.0625rem " ++ Themes.colorVariable Themes.SecondaryColor Colors.blueNordea)
+                        , outline none
+                        ]
+                    , Css.property "appearance" "none"
+                    , Css.property "-moz-appearance" "none"
+                    , Css.property "-webkit-appearance" "none"
```
fredrikloberg, line 213: Looks to be supported by all major browsers without the prefix: https://caniuse.com/mdn-css_properties_appearance_none
```suggestion
```

src/Nordea/Components/Checkbox.elm
```diff
@@ -222,11 +223,13 @@ view attrs (Checkbox config) =
...
             , appearanceStyle
             ]
-            :: attrs
+         ]
+            ++ attrs
         )
         [ Html.input
             [ type_ "checkbox"
             , name config.name
+            , Attrs.id config.name
```
fredrikloberg, line 232: An issue with using the name directly as the id is that the name in some cases should _not_ be unique. The name is required to be the same for groups of inputs, like radio buttons and checkboxes, etc., for the browser to know that they are semantically connected.

It can be solved by adding it as a required separate parameter, and provide it manually for every input. But the problem is avoided altogether if we make the input a child of the label instead.

fredrikloberg: As per the spec, it doesn't mention anything about using both nesting and `for` as far as I can tell (but the words `either` and `or`). So not sure if it is valid 🤔

https://www.w3.org/TR/html401/interact/forms.html#h-17.9.1

Screenshot 2022-10-27 at 11 17 54

src/Nordea/Components/MultiSelectDropdown.elm
```diff
@@ -170,7 +170,7 @@ view attrs dropdown =
     in
     Html.column attrs
         [ Html.row [ css [ displayFlex, justifyContent spaceBetween, marginBottom (rem 0.2) ] ]
-            [ Label.init { label = dropdown.label } |> Label.view []
+            [ Label.init { label = dropdown.label } |> Label.withAsLegend |> Label.view []
```
fredrikloberg, line 173: I suspect a `<fieldset>` tag is required as a wrapper for `legend` to be semantically connected 🤔

src/Nordea/Components/DropdownOptions.elm
vetlesolgaard: Good idea💡 It looks like the usage of `<summary>` is mainly intended for accordions with text. Let's keep what we have for now and we'll change it later if it is needed🙏

fredrikloberg: Ah yes, that may be true. This seems like a good resource for making it accessible: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-multiselectable

Both the example with a ul and li (with aria props) and the one with label and select look very interesting!

src/Nordea/Components/Checkbox.elm
vetlesolgaard: Good point! We'll add it for `input`s and `label`s on the same level🙏

> problem is avoided altogether if we make the input a child of the label instead

- Every major component library I find have labels and inputs on the same level in the DOM ([NAV](https://aksel.nav.no/designsystem/komponenter/textarea), [Entur](https://design.entur.org/komponenter/skjemaelementer/textfield), [Vy](https://spor.cloud.vy.no/komponenter/inputfelt))
- Documentation on w3's pages explicitly recommend to [use explicit labelling](https://www.w3.org/WAI/tutorials/forms/labels/#associating-labels-implicitly#associating-labels-implicitly)
- Having the `input` and the `label` on the same level in the DOM makes the elm component library so much easier to use for us as programmers.

I can't really see any reasons why we should stick to having the input as a child of the label.

src/Nordea/Components/MultiSelectDropdown.elm
vetlesolgaard: Good catch! I'll add it 🙌

src/Nordea/Components/Checkbox.elm
fredrikloberg: > Every major component library I find have labels and inputs on the same level in the DOM ([NAV](https://aksel.nav.no/designsystem/komponenter/textarea), [Entur](https://design.entur.org/komponenter/skjemaelementer/textfield), [Vy](https://spor.cloud.vy.no/komponenter/inputfelt))

The Vy label is not connected to the input, thats a bug 😛

> Having the input and the label on the same level in the DOM makes the elm component library so much easier to use for us as programmers.

The only difference I see is how the DOM ends up:

```html
<div>
    <label for="the-input">the label</label>
    <input id="the-input" />
</div>
```

vs:
```html
<label>
    the label
    <input />
</label>
```
The Elm interface can be identical (altough the first requires an extra param, the id), and both are adhering to the spec. The first one also requires a wrapping div in Elm (unless we make our components return lists instead of single tags), and the id handling adds quite a lot to complexity.



I see your points. My concern is only about the added complexity of the id handling. Is that added complexity worth it when both are adhering to the spec anyway? If it were React, I would be ok with both as the id handling can be totally hidden.

src/Nordea/Components/MultiSelectDropdown.elm
```diff
@@ -0,0 +1,250 @@
...
+        [ Html.row [ css [ displayFlex, justifyContent spaceBetween, marginBottom (rem 0.2) ] ]
+            [ Label.init { label = dropdown.label } |> Label.withAsLegend |> Label.view []
+            , dropdown.requirednessHint |> Html.viewMaybe RequirednessHint.view
+            ]
+        , Html.div
+            [ class "nordea-select"
+            , Events.on "focusout" (Decode.succeed (dropdown.onFocus False))
+            , Events.on "focusin" (Decode.succeed (dropdown.onFocus True))
+            , css [ position relative ]
+            , tabindex -1
```
fredrikloberg, line 181: I think these should be moved up to the root element. The expected behaviour for clicking labels is to transfer the focus to the form control. What do you think?

fredrikloberg: Also, I was unable to trigger the dropdown with just my keyboard, I think it may because of the `tabindex -1`. I think setting it to 0 is the way to go:


_tabindex="0" means that the element should be focusable in sequential keyboard navigation, after any positive tabindex values and its order is defined by the document's source order._

vetlesolgaard:
> A little design nitpick: the dropdown doesn't seem to be as wide as the collapsed view: Screenshot 2022-10-26 at 14 59 47
> 
> Looks to be due to the blue outline

It is the focus that styles it. It doesn't seem to be thought about in the component library. I'll remove the styling for the focus and add it later if needed😊

src/Nordea/Components/MultiSelectDropdown.elm
vetlesolgaard: Nice! That fixed it🙏 Not able to select or deselect checkboxes though, but that seems to be a problem with that component

fredrikloberg: Ah yes, just tested in PartnerHub. Tabbing checkboxes and radio buttons work, and you can toggle the state with space (as expected), but the focus outline doesn't show; making it look like it doesnt work. Thats a bug

src/Nordea/Components/Checkbox.elm
vetlesolgaard: Valid point with the complexity. On the other hand I think the components themselves will become less complex as we won't have to have a `Label` component needing to take into account what it contains. This will become much less complex I think.
I also feel a need to follow the recommendations made by [WAI](https://www.w3.org/WAI/tutorials/forms/labels/#associating-labels-implicitly#associating-labels-implicitly:~:text=but%20the%20labels.-,Associating%20labels%20implicitly,-In%20some%20situations).
Semantically it is also undoubtedly more correct with them on the same level as the input is not really "part of" the label.

fredrikloberg: I don't agree, but thats ok 🙂

I'm curious about your perspective on the added complexity of the nesting. The way I see it, there is a reduction of complexity with the nested variant. The div is just replaced with a label, we still need to take into account what it contains.

For/id: 
```elm
view props =
    Html.div
        [ css [ someLayoutCss ]]
        [ Html.label [ for props.inputId ] [ Text.bodyTextHeavy |> Text.view [] [ Html.text "Some label" ] ]
        , Html.input [ id props.inputId ] []
        ]
```

Nesting:
```elm
view props =
    Html.label
        [ css [ someLayoutCss ]]
        [ Text.bodyTextHeavy |> Text.view [] [ Html.text "Some label" ]
        , Html.input [] []
        ]
```

This is about whether we want to pay the price of id handling, or be pragmatic. The recommendation is also not about semantics, but screen reader support. It may be a valid point, but I dont know. What if it is based on older outdated screen readers that dont follow the spec; still worth it?

vetlesolgaard: We still need to take into account what it contains, but the `Label` component won't be the one having to support everything that can be contained.
I also think it is a problem that we have various components inside our `Label` component and not just the `Html.input`. Which makes the html itself harder to understand. How does a screenreader know which one of the texts inside the `Html.label` element is the actual label?
```elm
Html.label []
    [ Html.div [] [ Html.span [] [ "label" ], Html.span [] [ "requiredness" ]]
    , Html.input [] []
    , Html.div [] [ Html.span [] [ "error" ], Html.span [] [ "counter" ]
    ]
```

> This is about whether we want to pay the price of id handling, or be pragmatic

The price of id handling – is one extra property where needed which I don't think is a cost weighed up against all the benefits (IMO) of doing it this way. Either if we choose to enforce it as mandatory or with a function `withId : String -> Component -> Component`.

Other:
Some guy writes about (grain of salt) why nesting `input` inside `label` isn't supported for older screenreaders. Seems like the solution is still to connect them with `for`, `id` attributes. ([Link](https://dev.to/thejaredwilcurt/the-truth-behind-implicitexplicit-form-labels-2e10))

vetlesolgaard: We can take it up in technical town hall and see if anyone has any input. We'll go for the option which most think is the reasonable one😊 (We don't need to make a decision in the meeting itself)

PR #251, opened 2022-10-31
PAR-1878: Prevents event from selecting checkbox from bubbling up

vetlesolgaard:
### What was implemented? ✨🌈
- Removed default styling from `fieldset` and `ul` html elements. Default styling were not visible here for some reason. Could be css files added by explorer library? 
- Also prevents event from bubbling up when selecting an option in the dropdown

### DIFF
--- a/src/Nordea/Components/MultiSelectDropdown.elm
+++ b/src/Nordea/Components/MultiSelectDropdown.elm
@@ -31,18 +31,23 @@ import Css
         , cursor
         , display
         , displayFlex
+        , fitContent
         , flexDirection
         , fontSize
         , height
         , hover
         , inherit
+        , int
         , justifyContent
         , left
         , lineHeight
         , listStyle
         , marginBottom
+        , marginInlineEnd
+        , marginInlineStart
         , maxHeight
         , maxWidth
+        , minWidth
         , none
         , overflowY
         , padding
@@ -60,12 +65,13 @@ import Css
         , spaceBetween
         , top
         , transparent
-        , width
+        , zIndex
         )
 import Html.Styled as Html exposing (Attribute, Html)
-import Html.Styled.Attributes exposing (class, css, name, tabindex)
+import Html.Styled.Attributes exposing (css, name, tabindex)
 import Html.Styled.Events as Events
 import Json.Decode as Decode
+import Maybe.Extra as Maybe
 import Nordea.Components.Checkbox as Checkbox
 import Nordea.Components.Util.Hint as Hint
 import Nordea.Components.Util.Label as Label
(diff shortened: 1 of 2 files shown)


src/Nordea/Components/MultiSelectDropdown.elm
```diff
@@ -160,37 +178,48 @@ view attrs dropdown =
...
+            [ minWidth fitContent
+            , zIndex (int 1)
+            , displayFlex
+            , flexDirection column
+            , marginInlineStart (rem 0)
+            , marginInlineEnd (rem 0)
+            , Css.property "padding-block-start" "0"
+            , Css.property "padding-block-end" "0"
+            , Css.property "padding-inline-start" "0"
+            , Css.property "padding-inline-end" "0"
```
fredrikloberg, line 203: I think this should be enough; setting the shorthand should set all sub-variants too
```suggestion
            , margin (rem 0) 
            , padding (rem 0)
```

src/Nordea/Components/MultiSelectDropdown.elm
```diff
@@ -160,37 +178,48 @@ view attrs dropdown =
...
+            , Css.property "border" "none"
+            , Css.property "min-inline-size" "min-content"
```
fredrikloberg, line 205: Is this needed when we have minWidth?
```suggestion
```

src/Nordea/Components/MultiSelectDropdown.elm
```diff
@@ -160,37 +178,48 @@ view attrs dropdown =
                         , maxHeight (rem 16.75)
                         , dropdownStyles
                         , listStyle none
+                        , Css.property "margin-block-start" "0"
+                        , Css.property "margin-block-end" "0"
+                        , Css.property "padding-inline-start" "0"
```
fredrikloberg, line 183:
```suggestion
                        , margin (rem 0)
                        , padding (rem 0)
```

src/Nordea/Components/MultiSelectDropdown.elm
```diff
@@ -160,37 +178,48 @@ view attrs dropdown =
...
         [ Html.row [ css [ displayFlex, justifyContent spaceBetween, marginBottom (rem 0.2) ] ]
             [ Label.init { label = dropdown.label } |> Label.withAsLegend |> Label.view []
             , dropdown.requirednessHint |> Html.viewMaybe RequirednessHint.view
             ]
-        , Html.div
-            [ class "nordea-select"
-            , css [ position relative ]
-            ]
+            |> Html.showIf (isLabel || isRequirednessHint)
+        , Html.div [ css [ position relative ] ]
```
fredrikloberg, line 215: Do we need this div, or can the css be moved up to the fieldset and this div deleted?

fredrikloberg (approved):
Just some nitpicks, else looks good

fredrikloberg:
> * Removed default styling from `fieldset` and `ul` html elements. Default styling were not visible here for some reason. Could be css files added by explorer library?

Yes, the explorer library adds a lot of global css, affecting a lot 😕

src/Nordea/Components/MultiSelectDropdown.elm
vetlesolgaard: Deleted! Nice suggestion🙌

src/Nordea/Components/MultiSelectDropdown.elm
vetlesolgaard: Doesn't look like it 😊

src/Nordea/Components/MultiSelectDropdown.elm
vetlesolgaard: Also needed to set this on `ul`, but other than that it worked with setting margin and padding to 0🙏

PR #255, opened 2022-11-18
PAR-2020: Added FilterLabel component

vetlesolgaard:
Added component that is intended to work together with the `MultiSelectDropdown` component. The filter label will show when that specific filter is active in `MultiSelectDropdown`, and it will be possible to remove the filter when clicking the "cross" icon on `FilterLabel` component

Screenshot 2022-11-18 at 10 42 50

### DIFF
--- a/src/Nordea/Components/FilterLabel.elm
+++ b/src/Nordea/Components/FilterLabel.elm
@@ -1,0 +1,67 @@
+module Nordea.Components.FilterLabel exposing (init, view)
+
+import Css
+    exposing
+        ( alignSelf
+        , border3
+        , borderBox
+        , borderRadius
+        , boxSizing
+        , center
+        , cursor
+        , displayFlex
+        , fitContent
+        , flexDirection
+        , fontFamilies
+        , maxWidth
+        , noWrap
+        , padding2
+        , pointer
+        , rem
+        , solid
+        , whiteSpace
+        , width
+        )
+import Html.Styled as Html exposing (Attribute, Html)
+import Html.Styled.Attributes exposing (css)
+import Nordea.Components.Text as Text
+import Nordea.Resources.Colors as Colors
+import Nordea.Resources.Icons as Icon
+import Nordea.Themes as Themes
+
+
+type FilterLabel
+    = FilterLabel { label : String }
+
+
+init : { label : String } -> FilterLabel
+init { label } =
+    FilterLabel { label = label }
+
+
+view : List (Attribute msg) -> FilterLabel -> Html msg
+view attrs (FilterLabel { label }) =
+    Html.button
+        ([ css
+            [ padding2 (rem 0.25) (rem 0.75)
+            , displayFlex
+            , flexDirection Css.row
+            , borderRadius (rem 0.5)
+            , fontFamilies [ "inherit" ]
+            , borderRadius (rem 2)
+            , cursor pointer
+            , boxSizing borderBox
+            , Themes.backgroundColor Themes.SecondaryColor Colors.cloudBlue
+            , Themes.color Themes.PrimaryColor Colors.deepBlue
+            , border3 (rem 0.125) solid Colors.transparent
+            , maxWidth fitContent
+            , Css.property "gap" "1rem"
+            , whiteSpace noWrap
+            , alignSelf center
+            ]
+         ]
+            ++ attrs
+        )
+        [ Text.bodyTextSmall |> Text.view [] [ Html.text label ]
+        , Icon.cross [ css [ width (rem 1.0) ] ]
+        ]
--- a/README.md
+++ b/README.md
@@ -18,4 +18,4 @@ We currently have three css variables for themeing: `PrimaryColor`, `PrimaryColo
 * blueNordea -> PrimaryColorLight
 * blueMedium -> SecondaryColor
 * blueHaas -> SecondaryColor
-* blueCloud -> SecondaryColor
+* cloudBlue -> SecondaryColor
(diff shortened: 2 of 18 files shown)


src/Nordea/Components/FilterLabel.elm
```diff
@@ -0,0 +1,64 @@
...
+            , maxWidth fitContent
+            , Css.property "gap" "1rem"
+            , whiteSpace noWrap
+            , alignSelf center
+            ]
+         ]
+            ++ attrs
+        )
+        [ Text.bodyTextSmall |> Text.view [] [ Html.text label ]
+        , Icon.cross [ css [ width (rem 1.0) ] ]
```
fredrikloberg, line 63: Should probably be a button tag somewhere here for accessibility. I'd say either just around the cross icon if only that should be clickable, or the whole component if the whole should be clickable

src/Nordea/Components/FilterLabel.elm
```diff
@@ -0,0 +1,67 @@
...
+init { label } =
+    FilterLabel { label = label }
+
+
+view : List (Attribute msg) -> FilterLabel -> Html msg
+view attrs (FilterLabel { label }) =
+    Html.button
+        ([ css
+            [ padding2 (rem 0.25) (rem 0.75)
+            , displayFlex
```
fredrikloberg, line 47: Did you test in safari or firefox? I suspect we might need a `appearance: none` to get rid of the default button styles in one of those browsers

src/Nordea/Components/FilterLabel.elm
```diff
@@ -0,0 +1,67 @@
...
+            , flexDirection Css.row
+            , borderRadius (rem 0.5)
+            , fontFamilies [ "inherit" ]
+            , borderRadius (rem 2)
+            , cursor pointer
```
fredrikloberg, line 52: Not needed for button?
```suggestion
```

src/Nordea/Components/FilterLabel.elm
```diff
@@ -0,0 +1,67 @@
...
+            , boxSizing borderBox
+            , Themes.backgroundColor Themes.SecondaryColor Colors.cloudBlue
+            , Themes.color Themes.PrimaryColor Colors.deepBlue
+            , border3 (rem 0.125) solid Colors.transparent
+            , maxWidth fitContent
+            , Css.property "gap" "1rem"
+            , whiteSpace noWrap
+            , alignSelf center
```
fredrikloberg, line 60: I think it may be better if this is up the the parent and the context it is being used
```suggestion
```

src/Nordea/Components/FilterLabel.elm
```diff
@@ -0,0 +1,67 @@
...
+            , displayFlex
+            , flexDirection Css.row
+            , borderRadius (rem 0.5)
+            , fontFamilies [ "inherit" ]
+            , borderRadius (rem 2)
+            , cursor pointer
+            , boxSizing borderBox
+            , Themes.backgroundColor Themes.SecondaryColor Colors.cloudBlue
+            , Themes.color Themes.PrimaryColor Colors.deepBlue
+            , border3 (rem 0.125) solid Colors.transparent
```
fredrikloberg, line 56: Since it is transparent anyway
```suggestion
            , borderWidth (rem 0)
```

src/Nordea/Components/FilterLabel.elm
```diff
@@ -0,0 +1,67 @@
...
+view : List (Attribute msg) -> FilterLabel -> Html msg
+view attrs (FilterLabel { label }) =
+    Html.button
+        ([ css
+            [ padding2 (rem 0.25) (rem 0.75)
+            , displayFlex
+            , flexDirection Css.row
+            , borderRadius (rem 0.5)
+            , fontFamilies [ "inherit" ]
+            , borderRadius (rem 2)
```
fredrikloberg, line 51: Also, is the fontFamilies needed?

```suggestion
            , fontFamilies [ "inherit" ]
            , borderRadius (rem 2)
```

fredrikloberg (approved):
Looks good!

src/Nordea/Components/FilterLabel.elm
vetlesolgaard: Tested now. Button is looking fine. The only styling they add is `onClick`. Is that something we want to remove? 🤔

src/Nordea/Components/FilterLabel.elm
vetlesolgaard: Changed to `alignItems center`, that work? 🤌

src/Nordea/Components/FilterLabel.elm
vetlesolgaard: Good catch! Is it something we have to set at all? 🤔

src/Nordea/Components/FilterLabel.elm
fredrikloberg: If that is the intention, it is aligning the children vs aligning itself in its parent container

src/Nordea/Components/FilterLabel.elm
fredrikloberg: I think _border width 0_ and _appearance none_ are two ways to remove the default styling from buttons. Some browsers prefer one over the other, so having both is common; if I remember correctly.

src/Nordea/Components/FilterLabel.elm
vetlesolgaard: Yeah, the intention was to align children. Not the component itself 🙏

PR #259, opened 2022-12-02
Delete exposed 'withIsDisabled' in NumberInput component

wenwei-lai:
To disable the NumberInput component, we should use e.g. `|> NumberInput.view [ disabled True ]`

Deleted unused `withIsDisabled` to avoid confusing

### DIFF
--- a/src/Nordea/Components/NumberInput.elm
+++ b/src/Nordea/Components/NumberInput.elm
@@ -156,6 +156,7 @@ getAttributes config =
         , config.placeholder |> Maybe.map placeholder
         , config.onInput |> Maybe.map onInput
         , config.onBlur |> Maybe.map onBlur
+        , Just config.isDisabled |> Maybe.map Attributes.disabled
         , Just config.value |> Maybe.map format |> Maybe.map value
         ]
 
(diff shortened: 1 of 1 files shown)


src/Nordea/Components/NumberInput.elm
```diff
@@ -156,6 +156,7 @@ getAttributes config =
         , config.placeholder |> Maybe.map placeholder
         , config.onInput |> Maybe.map onInput
         , config.onBlur |> Maybe.map onBlur
+        , Just config.isDisabled |> Maybe.map Attributes.disabled
```
fredrikloberg, line 159: Instead, I suggest to remove it from the config. It is already supported to set disabled as a normal attribute:

`|> NumberInput.view [ disabled True ]`

This is also the interface we use for disabled for the other components as well

```suggestion
```

wenwei-lai: @fredrikloberg Good point! Do you think we should then remove the exposed `withIsDisabled` ? 🤔If that is not the right way to do disabling

```
withIsDisabled : Bool -> NumberInput msg -> NumberInput msg
withIsDisabled val (NumberInput config) =
    NumberInput { config | isDisabled = val }
```

fredrikloberg: Yea 👍

wenwei-lai: @fredrikloberg Great 😍 and done! Could you please take one more look at it ?

PR #262, opened 2022-12-06
PAR-1921: Added fonts with new format. Removed unused fonts

vetlesolgaard:
- Removed unused fonts 
- Added `fonts.css` file that can be loaded by other applications that is using the component library. This way these applications won't have to hold their own typography. 

[See connected PR in PartnerHub](https://github.com/SGFinansAS/PartnerHub/pull/1744)


fredrikloberg (approved):
Noice, looks good!

PR #263, opened 2022-12-07
PAR-2056: Added contrast to text + refactor

vetlesolgaard:
- Added required contrast to text inside upload component
- Refactored a bit


fredrikloberg (approved):
🙌

PR #270, opened 2023-01-10
NumberInput formatter not works for typing decimal

wenwei-lai:
The formatter of NumberInput component prevents users from typing decimal because the formatter would format unfinished number typing.

e.g. when users try to type "123.45", before finish typing, the formatter would format "123." to "123". Which make it not possible to type decimal numbers.

### DIFF
--- a/src/Nordea/Components/NumberInput.elm
+++ b/src/Nordea/Components/NumberInput.elm
@@ -126,7 +126,7 @@ view attributes (NumberInput config) =
 getAttributes : Config msg -> List (Attribute msg)
 getAttributes config =
     let
-        format: String -> String
+        format : String -> String
         format value =
             if String.endsWith "." value || String.endsWith "," value then
                 value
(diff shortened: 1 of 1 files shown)


src/Nordea/Components/NumberInput.elm
```diff
@@ -127,8 +127,8 @@ getAttributes : Config msg -> List (Attribute msg)
 getAttributes config =
     let
         format value =
-            if value == "" then
-                ""
+            if String.endsWith "." value || String.endsWith "," value then
+                value
```
wenwei-lai, line 131:
```
                if value == "" then ""
```
This looks redundant because value = "" would return "" as well if it goes into the `else` track

Rostislav013 (approved):

src/Nordea/Components/NumberInput.elm
perok: I guess it avoids the toFloat computation that is guaranteed failure?

perok (approved):
:+1:

src/Nordea/Components/NumberInput.elm
wenwei-lai: Yeah, i think so. The purpose might be to avoid that. But it actually does not matter because `else` track also handle well the situation if toFloat fails

PR #271, opened 2023-01-17
Extra small table variant

trysnetron:
## Screenshot 📷 
bilde

## Features ✨ 
1. Adds a smaller table variant (FinansFront ❤️)
2. Exposes the internal variant type instead of creating `tbodyExtraSmall` and `theadExtraSmall` functions
3. Explorer example
4. Changes the `t(head|body)WithVariant` function API to take `Variant` first, a little bit cleaner IMO 

## Discussion points ⚔️ 

Should we remove the `t(head|body)Small` functions? They make writing tables a little bit terser, but I would argue that fidgeting with the `table`, `thead` and `tbody` elements in the HTML is very "low level" and will probably be abstracted away anyway by to some nice function like `myNiceTableFunc: List MyCoolDomainModelType -> Html msg`. This _will_ break backwards compatibility though so is it worth it to have a "neater" `Table` component API? 

@vetlesolgaard would appreciate you and/or your team's input on this 😊 🙏

### DIFF
--- a/src/Nordea/Components/Table.elm
+++ b/src/Nordea/Components/Table.elm
@@ -7,6 +7,9 @@ module Nordea.Components.Table exposing
     , theadSmall
     , tr
     , view
+    , Variant
+    , theadWithVariant
+    , tbodyWithVariant
     )
 
 import Css
--- a/src/Nordea/Components/Table.elm
+++ b/src/Nordea/Components/Table.elm
@@ -73,7 +73,7 @@ theadWithVariant attrs children variant =
                     rem 3.5
 
                 ExtraSmall ->
-                    rem 1.5
+                    rem 2.5
     in
     Html.thead
         (css
@@ -136,7 +136,7 @@ tbodyWithVariant attrs children variant =
                     rem 4.5
 
                 ExtraSmall ->
-                    rem 2.5
+                    rem 1.5
     in
     Html.tbody
         (css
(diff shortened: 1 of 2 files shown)


src/Nordea/Components/Table.elm
```diff
@@ -57,8 +61,8 @@ view attrs children =
 
 {-| Groups the header content in a table
 -}
-theadWithVariant : List (Attribute msg) -> List (Html msg) -> Variant -> Html msg
-theadWithVariant attrs children variant =
+theadWithVariant : Variant -> List (Attribute msg) -> List (Html msg) -> Html msg
+theadWithVariant variant attrs =
```
Eivinddh, line 65: I guess one thing to consider is if you want to write code like:
```
ExtraSmall
|> theadWithVariant [] [ Html.text ""]
```
This will not work, but I'm not certain if its nice

trysnetron: Discussed this with @fredrikloberg now and I think having `Variant` as the first argument is fine for now. 

We are using the [builder pattern](https://sporto.github.io/elm-patterns/basic/builder-pattern.html) for a lot of other components and there it makes sense to pass in `Variant` as the last argument, but it is not the case for the functions in the Table component.

If we were to improve the Table component in the future and wanted to have optional parameters, we could then either go for the builder pattern or just an [argument list](https://sporto.github.io/elm-patterns/basic/arguments-list.html) which might be neater (and is already what is used in i.e. the Elm Html package).

fredrikloberg (approved):

PR #277, opened 2023-02-08
Add cross to InfoBanner and some updates

fonnes:
- Rename to infobanner
- Use config rather than builder pattern
- Deleted openable because it was not used

image
image


src/Nordea/Components/InfoBanner.elm
```diff
@@ -0,0 +1,85 @@
...
+import Nordea.Resources.Icons as Icons
+import Nordea.Themes as Themes
+
+
+type Appearance
+    = Warning
+    | Info
+
+
+type Config msg
```
morteako, line 36: do we need the type parameter here?

fonnes: Nop! :D

src/Nordea/Components/InfoBanner.elm
```diff
@@ -0,0 +1,85 @@
...
+    = Appearance Appearance
+
+
+view : List (Config msg) -> List (Attribute msg) -> List (Html msg) -> Html msg
```
morteako: should we really allow a list when only the first argument is used?
could use Maybe?

src/Nordea/Components/InfoBanner.elm
```diff
@@ -0,0 +1,85 @@
...
+view config attrs children =
+    let
+        appearance =
+            config
+                |> List.filterMap
```
morteako, line 45: could just be normal List.map i think?

src/Nordea/Components/InfoBanner.elm
fonnes: Enforcing new pattern where you provide a list of optional arguments instead of the verbose builder pattern

src/Nordea/Components/InfoBanner.elm
fonnes: Trying to be an example of the new pattern

tbnorheim69 (commented):
Looks okey :)

fredrikloberg (approved):

PR #278, opened 2023-02-13
Implement AccordionMenu

2pacalypse:


fredrikloberg (commented):
Nice work! 🤩

I have some suggestions you may consider:

1. Semantics and accessibility
I think a semantic way to create an accordion is to use the `details` and `summary` tags. See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details

```html
<details>
  <summary>
    <!-- Label for the disclosure widget -->
  </summary>
  <div> <!-- optional wrapper (for styling) -->
    <!-- collapsible content goes here -->
  </div>
</details>
```

2. Making the component more generic
I think it is beneficial to make the component as generic and reusable as possible, so that it can be used by the other teams as well with minimum changes required. I suggest an interface like this for the content:

```elm
AccordionMenu.init 
    model.customModel.accordionMenuIsOpen 
    ToggleAccordionMenu
    |> AccordionMeny.view []
            [ AccordionMenu.header [] [ {-Any custom html goes here-} ]
            , AccordionMenu.content [] [ {-Any custom html goes here-} ]
            ]
```

3. Moving away from the builder pattern
We have just started to move away from the builder pattern in the component library. So if you want, you may opt for the new pattern for this component too. See https://github.com/SGFinansAS/elm-components/pull/277 for an example. With the new pattern, the interface may look something like this:
```elm
AccordionMenu.view 
    { isOpen = model.customModel.accordionMenuIsOpen, onToggle = ToggleAccordionMenu }
    [] -- <- Optional component specific props
    [] -- <- Html / elm native attributes (css etc.)
    [ AccordionMenu.header [] [ {-Any custom html goes here-} ]
    , AccordionMenu.content [] [ {-Any custom html goes here-} ]
    ]
```

fredrikloberg (commented):
I see that we already have a component called Accordion, but that one kinda looks like a FAQ entry to me 🤔 I'm tempted to suggest to rename the existing one to Faq or something, and then call this new one just Accordion. Do you have any thoughts? May also be good to discuss the naming of these components with Line as well

2pacalypse:
Thanks, I will reflect these into the code

explorer/src/Stories/AccordionMenu.elm
```diff
@@ -0,0 +1,74 @@
...
+                    [ AccordionMenu.header [] [ text "Kontrakt e-signert 10.10.22" ]
+                    , AccordionMenu.content []
+                        [ ul []
+                            [ li [] [ text "Undertegner identifisert 09.10.22" ]
+                            , li [] [ text "E-signering utloper snart 08.10.22" ]
+                            , li [] [ text "E-signering sendt 07.10.22" ]
+                            ]
+                        ]
+                    ]
+          , {}
```
2pacalypse, line 32: I have followed the structure you proposed but instead of a record for required arguments, I opted for optional arguments for everything including the initial state.

src/Nordea/Components/AccordionMenu.elm
```diff
@@ -0,0 +1,63 @@
...
+                    attrs
+
+        attrs_ =
+            css
+                [ children [ typeSelector "summary" [ listStyle none, children [ typeSelector "div" [ lastChild [ display inlineBlock ], nthLastChild "2" [ display none ] ] ] ] ]
+                , withAttribute "open" [ children [ typeSelector "summary" [ children [ typeSelector "div" [ lastChild [ display none ], nthLastChild "2" [ display inlineBlock ] ] ] ] ] ]
+                ]
+                :: openAttr
+    in
+    Html.details attrs_ children_
```
2pacalypse, line 47: It turns out the HTML tags you mentioned already handle opening and closing of the accordion. I eliminated the message from the component then. The effort was in customizing the chevron by changing visibility of elements in 2 states of `<details>` and `details open`.

vetlesolgaard:
> I see that we already have a component called Accordion, but that one kinda looks like a FAQ entry to me 🤔 I'm tempted to suggest to rename the existing one to Faq or something, and then call this new one just Accordion. Do you have any thoughts? May also be good to discuss the naming of these components with Line as well

I think this should be discussed with Line. Should we have 2 different accordions? Are they used for the same purposes? Are they really just the same component on different backgrounds?

2pacalypse:
@fredrikloberg For the naming, you are right there is an overlap with Accordion component. Maybe we can call this a Details, DetailsSummary component. What do you think?

src/Nordea/Components/AccordionMenu.elm
```diff
@@ -0,0 +1,63 @@
...
+    = Open
+    | Closed
+
+
+type OptionalConfig
+    = Status Status
+
+
+view : List OptionalConfig -> List (Attribute msg) -> List (Html msg) -> Html msg
+view config attrs children_ =
```
vetlesolgaard, line 20: Since the functionality of if the accordion is open or not is so integral to how this component works, I think it would better be represented by a record where the value is mandatory rather than a list of optional arguments 😊 We could also consider representing it with a `Bool` since it will always be just open or closed. 
What do you think? 😊
```suggestion
type alias Config =
     { isOpen : Bool }

view : Config -> List (Attribute msg) -> List (Html msg) -> Html msg
view config attrs children_ =
```

src/Nordea/Components/AccordionMenu.elm
```diff
@@ -0,0 +1,63 @@
...
+
+
+header : List (Attribute msg) -> List (Html msg) -> Html msg
+header attrs children =
+    Html.summary
+        (css [ displayFlex, alignItems center, justifyContent spaceBetween ] :: attrs)
+        (List.append [ div [] children ]
+            [ div [ css [ float right ] ] [ Icons.chevronDown [ css [ width (rem 1.5) ] ] ]
+            , div [ css [ float right ] ] [ Icons.chevronLeft [ css [ width (rem 1.5) ] ] ]
+            ]
```
vetlesolgaard, line 57:
```suggestion
        [ div [] children
        , div [ css [ float right ] ] [ Icons.chevronDown [ css [ width (rem 1.5) ] ] ]
        , div [ css [ float right ] ] [ Icons.chevronLeft [ css [ width (rem 1.5) ] ] ]
        ]
```

src/Nordea/Components/AccordionMenu.elm
```diff
@@ -0,0 +1,63 @@
...
+                , withAttribute "open" [ children [ typeSelector "summary" [ children [ typeSelector "div" [ lastChild [ display none ], nthLastChild "2" [ display inlineBlock ] ] ] ] ] ]
+                ]
+                :: openAttr
+    in
+    Html.details attrs_ children_
+
+
+header : List (Attribute msg) -> List (Html msg) -> Html msg
+header attrs children =
+    Html.summary
```
vetlesolgaard, line 52: Should we also have some pointer styling when user is hovering over this element so they understand that it is clickable? 😊 (`cursor pointer`)

2pacalypse:
> > I see that we already have a component called Accordion, but that one kinda looks like a FAQ entry to me thinking I'm tempted to suggest to rename the existing one to Faq or something, and then call this new one just Accordion. Do you have any thoughts? May also be good to discuss the naming of these components with Line as well
> 
> I think this should be discussed with Line. Should we have 2 different accordions? Are they used for the same purposes? Are they really just the same component on different backgrounds?

I think they are the same. Initially, I tried to make use of Accordion and then it turned out I had to change some details in the code like the background color. Or the title was only expecting a string in the original Accordion and it was a problem for Line's sketch which showed icons next to titles. Then I wanted to re-do it with an influence from Accordion but I really like the `<details><summary>` usage in this version because it eliminates the need for message in the interface.

src/Nordea/Components/AccordionMenu.elm
2pacalypse: Yes, making it a mandatory field could be a better idea. I will do it.

src/Nordea/Components/AccordionMenu.elm
2pacalypse: In the last story, there is an example of `cursor pointer` but yes in the default usage it is not like that. I could do it.

src/Nordea/Components/AccordionMenu.elm
```diff
@@ -0,0 +1,63 @@
...
+                    attribute "open" "" :: attrs
+
+                Closed ->
+                    attrs
+
+        attrs_ =
+            css
+                [ children [ typeSelector "summary" [ listStyle none, children [ typeSelector "div" [ lastChild [ display inlineBlock ], nthLastChild "2" [ display none ] ] ] ] ]
+                , withAttribute "open" [ children [ typeSelector "summary" [ children [ typeSelector "div" [ lastChild [ display none ], nthLastChild "2" [ display inlineBlock ] ] ] ] ] ]
+                ]
```
fredrikloberg, line 44: May be easier to read with class names:
```suggestion
    css
        [ children [ typeSelector "summary" [ listStyle none, descendants [ class "left-arrow" [ display inlineBlock ], class "down-arrow" [ display none ] ] ] ]
        , withAttribute "open"
            [ children [ typeSelector "summary" [ descendants [ class "left-arrow" [ display none ], class "down-arrow" [ display inlineBlock ] ] ] ]
            ]
        ]
```

2pacalypse: I wanted to assign them a class like that but then in a use like this, if anyone actually uses two or three elements which belong to class `down-arrow` inside the `summary` tag, I thought it would be problematic. Are these names defined somewhere or can I come up with names like that?

fredrikloberg: We very rarely use class names in elm (only for special cases like these), so I don't think it will be a problem. Most css is also scoped through elm-css

src/Nordea/Components/AccordionMenu.elm
2pacalypse: nice one! I will revise.

vetlesolgaard:
> 

Could be good to have a quick chat with Line about this😊

vetlesolgaard:
> > > I see that we already have a component called Accordion, but that one kinda looks like a FAQ entry to me thinking I'm tempted to suggest to rename the existing one to Faq or something, and then call this new one just Accordion. Do you have any thoughts? May also be good to discuss the naming of these components with Line as well
> > 
> > 
> > I think this should be discussed with Line. Should we have 2 different accordions? Are they used for the same purposes? Are they really just the same component on different backgrounds?
> 
> I think they are the same. Initially, I tried to make use of Accordion and then it turned out I had to change some details in the code like the background color. Or the title was only expecting a string in the original Accordion and it was a problem for Line's sketch which showed icons next to titles. Then I wanted to re-do it with an influence from Accordion but I really like the `<details><summary>` usage in this version because it eliminates the need for message in the interface.

I'm a little bit confused on how this accordion should work in other use cases. Could be good to have a quick chat with Line about this😊

src/Nordea/Components/AccordionMenu.elm
```diff
@@ -0,0 +1,46 @@
...
+    Html.details attrs_ children_
+
+
+header : List (Attribute msg) -> List (Html msg) -> Html msg
+header attrs children =
+    Html.summary
+        (css [ displayFlex, alignItems center, justifyContent spaceBetween ] :: attrs)
+        [ div [] children
+        , div [ Html.Styled.Attributes.class "accordion-open-icon", css [ float right ] ] [ Icons.chevronDown [ css [ width (rem 1.5) ] ] ]
+        , div [ Html.Styled.Attributes.class "accordion-close-icon", css [ float right ] ] [ Icons.chevronUp [ css [ width (rem 1.5) ] ] ]
```
fredrikloberg, line 40: Are the floats and the wrapping divs needed?
```suggestion
        , Icons.chevronDown [ class "accordion-open-icon", css [ width (rem 1.5) ] ]
        , Icons.chevronUp [ class "accordion-close-icon", css [ width (rem 1.5) ] ]
```

fredrikloberg (approved):
Looks very good! 🚀 The only thing is the naming, but we may leave that for later

src/Nordea/Components/AccordionMenu.elm
2pacalypse: after using the class selector, I now see we can get rid of the divs. floats are also not needed as we use flex box with space between those 3 elements (header content + arrow + arrow), one of the arrows being hidden always. nice!

PR #289, opened 2023-03-17
Add SkeletonLoader component

hannabergljot:
Skeleton loader is already used in PH, so added here as a component. The component in itself is an empty div with styling, so shape and size specifications are passed as parameters. I added two examples to the UI Explorer for illustrations.

Screenshots:
Screenshot 2023-03-17 at 9 16 33 AM
Screenshot 2023-03-17 at 9 16 52 AM

### DIFF
--- a/src/Nordea/Components/SkeletonLoader.elm
+++ b/src/Nordea/Components/SkeletonLoader.elm
@@ -1,0 +1,42 @@
+module Nordea.Components.SkeletonLoader exposing (view)
+
+import Css
+    exposing
+        ( animationDuration
+        , animationName
+        , ms
+        )
+import Css.Animations exposing (keyframes)
+import Html.Styled as Html
+    exposing
+        ( Attribute
+        , Html
+        )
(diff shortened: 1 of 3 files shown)


Rostislav013 (approved):

fredrikloberg (approved):
🙌

wenwei-lai (approved):
:rocket:

src/Nordea/Components/SkeletonLoader.elm
```diff
@@ -0,0 +1,42 @@
...
+    exposing
+        ( Attribute
+        , Html
+        )
+import Html.Styled.Attributes exposing (css)
+
+
+view : List (Attribute msg) -> Html msg
+view attrs =
+    Html.div
```
fredrikloberg, line 20: Would be good for accessibility to add these attributes:
```elm
Attrs.attribute "aria-busy" "true"
Attrs.attribute "aria-live" "polite"
```

PR #293, opened 2023-04-11
PAR-2221: customize accordionmenu

fredrikloberg:
Small adjustments to the AccordionMenu:
* Remove wrapper div's to make it easier to style
* Minor formatting


src/Nordea/Components/AccordionMenu.elm
```diff
@@ -11,36 +26,58 @@ type alias Config =
...
             else
-                attrs
-
-        attrs_ =
-            css
-                [ children [ typeSelector "summary" [ listStyle none, descendants [ class "accordion-closed-icon" [ display inlineBlock ], class "accordion-open-icon" [ display none ] ] ] ]
-                , withAttribute "open" [ children [ typeSelector "summary" [ descendants [ class "accordion-closed-icon" [ display none ], class "accordion-open-icon" [ display inlineBlock ] ] ] ] ]
-                ]
-                :: openAttr
+                attribute "closed" ""
```
2pacalypse, line 37:
```suggestion
                fromUnstyled empty
```
I agree this section needs refactoring -- being my first ever contribution :sweat_smile: . I think using `closed` might give someone a false impression though such attribute exists on `Html.details` but I think all we have is `open`. Any other string would also do it but using the empty one seems more convenient?

fredrikloberg: Good idea! Updated now

2pacalypse (approved):

PR #298, opened 2023-04-21
PAR: Dropdown bugfix

fredrikloberg:
* The filled chevron icon for dropdowns were not working properly.
* some misc padding changes for consistency


src/Nordea/Html.elm
```diff
@@ -91,4 +91,4 @@ styleIf condition style =
         style
 
     else
-        Css.property "" ""
+        Css.property "_" "_"
```
fonnes, line 94: :O

PR #307, opened 2023-06-20
PAR-2452: Dropdown chevron icon size adjustment

heiditark:
Dropdown chevron icon size adjustment and padding change to consider also simple dropdown.


wenwei-lai (approved):
Cool stuff :sunglasses:  Just fix the CI issue:
```
(fix) NoUnused.Variables: Imported variable `paddingRight` is not used
```

PR #314, opened 2023-08-07
TextInput withRemoveInput feature

heiditark:
Added withRemoveInput feature to textInput component.

image


src/Nordea/Components/TextInput.elm
```diff
@@ -71,6 +73,8 @@ type alias Config msg =
     , maxLength : Maybe Int
     , pattern : Maybe String
     , hasSearchIcon : Bool
+    , hasRemoveInputButton : Bool
+    , onRemoveInput : Maybe msg
```
fredrikloberg, line 77: Is there a need to have both fields, or can the bool be inferred internally from the other? E.g., `hasRemoveInputButton = Maybe.isJust onRemoveInput`

fredrikloberg, line 77: What do you think about "clear" instead of "remove"?
```suggestion
    , hasClearInputButton : Bool
    , onClearInput : Maybe msg
```

heiditark: I think we need both fields since we can have some condition (for example no input typed) when we don't want to show the button. But once there is some input is typed we want to show the button.

src/Nordea/Components/TextInput.elm
```diff
@@ -161,7 +176,28 @@ onEnterPress msg =
...
+            )
+            [ styled a
+                [ backgroundColor Colors.transparent
+                , position absolute
+                , right (rem 0.35)
+                , top (rem 0.35)
+                , border3 (rem 0.125) solid Colors.transparent
+                ]
+                (Maybe.values [ config.onRemoveInput |> Maybe.map onClick ])
+                [ Icons.roundedCross [ css [ width (rem 1.5), color Colors.mediumGray ] ] ]
```
fredrikloberg, line 192: Do we need the wrapping a-tag, or can we put the onclick directly on the icon?
```suggestion
            [ Icons.roundedCross 
                [ Maybe.values [ config.onRemoveInput |> Maybe.map onClick ]
                , tabindex 0 -- remove if not needed. sometimes needed to make onclick work on divs.
                , css 
                    [ width (rem 1.5)
                    , color Colors.mediumGray
                    , position absolute
                    , right (rem 0.35)
                    , top (rem 0.35)
                    , padding (rem 1) -- should have some padding to increase clickable area for accessibility reasons. ideally, I'd say minimum 42x42px should be clickable
                    ]
                ]
```

src/Nordea/Components/TextInput.elm
fredrikloberg: Hmm, what if we standardized that behaviour and made it internal to this component? In other words, this component checks if the input is empty or not and shows/hides the button accordingly. Do you think that'll work?

heiditark: I'll try to implement that.

src/Nordea/Components/TextInput.elm
heiditark: I had to leave the wrapping a-tag but added the padding.

src/Nordea/Components/TextInput.elm
heiditark: I implemented it so that we're checking internally if onClearInput is Just and input is non-empty when showing the button.

src/Nordea/Components/TextInput.elm
fredrikloberg: What was the problem? I think an a-tag may be wrong in regards to accessibility as it is not linking anywhere, so a button tag may be more appropriate in case a wrapper is needed

fredrikloberg (approved):
Looks good! 💯

PR #315, opened 2023-08-17
Fix clipping of modal dialogs on narrow screens

atraastrum:
At least ESign is affected. What makes this particularly bad is inability to close the dialog for example on a mobile device if close button and eXit icon are on the right.

<details><summary>Old behavior GIF</summary>

clipped-modal

</details>

<details><summary>Fixed</summary>

fixed-modal

</details>

### DIFF
--- a/explorer/src/Stories/Modal.elm
+++ b/explorer/src/Stories/Modal.elm
@@ -1,7 +1,7 @@
 module Stories.Modal exposing (..)
 
 import Config exposing (Msg(..))
-import Css exposing (rem, width)
+import Css exposing (..)
 import Html.Styled as Html
 import Html.Styled.Attributes exposing (css)
 import Html.Styled.Events exposing (onClick)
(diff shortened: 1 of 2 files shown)


fredrikloberg:
I got a bit curious since it works in PartnerHub even without this change. The preview in the explorer is setup like this:
```elm
Modal.default ToggleModal
  |> Modal.withTitle "Title"
  |> Modal.view
      [ css [ width (rem 35) ] ]
      [ Html.text "Suggested width is 560px (35rem)" ]
```

In other words, the width is hardcoded to be 35rem. Based on this, I would expect the width to actually always be 35rem, and thus not be responsive either. If responsiveness is wanted, I would instead expect this css `css [ maxWidth (rem 35), width (pct 100) ]`. This is how it is used in PH, which explains why it is working.

What do you think?

atraastrum:
> I got a bit curious since it works in PartnerHub even without this change. The preview in the explorer is setup like this:
> 
> ```elm
> Modal.default ToggleModal
>   |> Modal.withTitle "Title"
>   |> Modal.view
>       [ css [ width (rem 35) ] ]
>       [ Html.text "Suggested width is 560px (35rem)" ]
> ```
> 
> In other words, the width is hardcoded to be 35rem. Based on this, I would expect the width to actually always be 35rem, and thus not be responsive either. If responsiveness is wanted, I would instead expect this css `css [ maxWidth (rem 35), width (pct 100) ]`. This is how it is used in PH, which explains why it is working.
> 
> What do you think?

I think I have made a mistake :) but when I try your suggestion in explorer the max-width is not applied:
image

Looks like it is because of this
image

So either I have to use `important` flag or somehow force elm to change the order.

Do you use `important` in PartnerHub?

fredrikloberg:
No we don't have important in PartnerHub for this, so probably just lucky with the ordering. The way it works is that elm-css hashes the content of the css block and then sorts it alphabetically, which results in a random order. No way to avoid this afaik, other that leveraging standard css precedence rules.

So what about removing the max and min-width from the element of the modal component? That should solve it

atraastrum:
> No we don't have important in PartnerHub for this, so probably just lucky with the ordering. The way it works is that elm-css hashes the content of the css block and then sorts it alphabetically, which results in a random order. No way to avoid this afaik, other that leveraging standard css precedence rules.
> 
> So what about removing the max and min-width from the element of the modal component? That should solve it

Applied your suggestion also added an example to explorer.

PR #320, opened 2023-10-02
add elm review simplify

morteako:
Add elm-review simplify and run all fixes.
Also run elm-json upgrade in /review

### DIFF
--- a/src/Nordea/Components/InfoBanner.elm
+++ b/src/Nordea/Components/InfoBanner.elm
@@ -42,13 +42,11 @@ view config attrs children =
     let
         appearance =
             config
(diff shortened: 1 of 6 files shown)


fredrikloberg (approved):
🙌

src/Nordea/Components/InfoBanner.elm
```diff
@@ -42,11 +42,11 @@ view config attrs children =
     let
         appearance =
             config
-                |> List.filterMap
+                |> List.map
```
morteako, line 45: Should do -> List.head and then
Maybe.map instead

PR #321, opened 2023-10-09
PAR-2347: Collapsible InformationDetails Card

heiditark:
Implemented collapsible Card in addition to the regular one. Collapsible card component is used in InformationDetails.  'emphasisedText' fades in and out when card is collapsed/uncollapsed.

image
image

### DIFF
--- a/src/Nordea/Components/InformationDetails.elm
+++ b/src/Nordea/Components/InformationDetails.elm
@@ -36,11 +36,11 @@ import Nordea.Resources.Colors as Colors
 
 card :
     List (Attribute msg)
-    -> Maybe String
     -> List (Html msg)
+    -> Maybe String
     -> Maybe { emphasisedText : Html msg, isOpen : Bool }
     -> Html msg
-card attrs title children collapsibleProps =
+card attrs children title collapsibleProps =
     let
         withOptionalTitle =
             Maybe.map Card.withTitle title
--- a/src/Nordea/Components/InformationDetails.elm
+++ b/src/Nordea/Components/InformationDetails.elm
@@ -1,6 +1,6 @@
 module Nordea.Components.InformationDetails exposing
     ( card
-    , collapsableCard
+    , collapsibleCard
     , element
     , fullWidthElement
     , label
@@ -82,7 +82,7 @@ card attrs children title =
             ]
 
 
-collapsableCard :
+collapsibleCard :
     { attrs : List (Attribute msg)
     , title : String
     , emphasisedText : String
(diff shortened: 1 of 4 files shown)


src/Nordea/Components/InformationDetails.elm
```diff
@@ -60,6 +82,66 @@ card attrs children title =
...
 
+collapsibleCard :
+    { attrs : List (Attribute msg)
+    , title : String
+    , emphasisedText : String
+    , isOpen : Bool
+    , children : List (Html msg)
+    }
+    -> Html msg
+collapsibleCard { attrs, title, emphasisedText, isOpen, children } =
```
heiditark, line 93: Should we instead have only one card and have it regular/collapsible based on `isCollapsible` parameter? This would be a small change but not sure if it's convenient since the collapsible card needs more/different parameters than regular one.

src/Nordea/Components/InformationDetails.elm
```diff
@@ -60,6 +82,66 @@ card attrs children title =
             ]
 
 
+collapsibleCard :
+    { attrs : List (Attribute msg)
+    , title : String
+    , emphasisedText : String
```
fredrikloberg, line 88: Instead of taking a string, what do you think about taking html? This makes it possible to slightly customize the content. For example, wrap the text in a span to change appearance, add accessibility attributes, or have a multiline text, etc.
```suggestion
    , emphasisedText : Html msg
```

src/Nordea/Components/InformationDetails.elm
fredrikloberg: I think it sounds like a good idea to merge the two. Regarding the parameters, what about taking an optional record?

```elm
card :
    { attrs : List (Attribute msg)
    , title : String
    , collapsibleProps: Maybe { emphasisedText : Html msg, isOpen: Bool }
    , children : List (Html msg)
    }
```

fredrikloberg (approved):
Great work!

src/Nordea/Components/InformationDetails.elm
heiditark: Do we still want the emphasisedText to have the fade in/out transition in all cases or should that be passed within the emphasisedText : Html msg?

fredrikloberg: I would say let it always have it. We can add a parameter for it later if needed

src/Nordea/Components/InformationDetails.elm
heiditark: I'm thinking how to do merge safely. Should we do the merge but leave old and new one and start slowly using the merged one in PH? Do you know if InformationDetails regular card is used outside of PartnerHub?

fredrikloberg: I would say do the merge in one go. We don't have to start using the collapsible feature all over at first, but we can set it to Nothing and get the old behaviour. I think they may also be used in the customer portal, maybe esign. But it shouldn't be a problem, they can set the prop to Nothing as well when they update.

heiditark: Yeah okay. I'll do that 👍

PR #322, opened 2023-10-12
PAR-2644: Small label adjustment

fredrikloberg:
Small label adjustment

### DIFF
--- a/src/Nordea/Components/Label.elm
+++ b/src/Nordea/Components/Label.elm
@@ -35,7 +35,6 @@ import Css
         , pseudoClass
         , rem
         , spaceBetween
-        , width
         , wrap
         )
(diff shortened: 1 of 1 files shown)


atraastrum (commented):
What does this fix?

fredrikloberg:
> What does this fix? I guess PAR-2644 is a ticket number in some bug tracking system? How can I see that ticket?

It improves column width behaviour when used inside a table layout. The change has no effect outside of that, but it makes more sense logically to think about it having `flexDirection column` rather than `flexDirection row` where all elements are wrapping anyway. The code is referring to a task in our [jira board](https://sgfinans.atlassian.net/jira/software/projects/PAR/boards/1?selectedIssue=PAR-2644)

atraastrum (approved):

PR #324, opened 2023-10-18
PAR-2624: use the onInput msg for onClear in text inputs

2pacalypse:
Eliminates the `if-else` in the `view`. e.g.

```
if Maybe.isJust config.onClearInput then
...
else if config.hasSearchIcon then
...
else
...

```
I think one could not have both cross and search icon at the same time previously due to that. Fixes to one single view.

Uses the webkit property for cross icon. This way, we can use the same onInput msg on clear. not available in firefox.

Screenshots from chrome
image
image

This one is a bit problematic. Do we like the placement of the cross, should we bring it left always or when there is currency or leave it like this? 
image

### DIFF
--- a/src/Nordea/Components/TextInput.elm
+++ b/src/Nordea/Components/TextInput.elm
@@ -53,7 +53,15 @@ import Css
         , width
         )
 import Html.Styled as Html exposing (Attribute, Html, input, styled)
-import Html.Styled.Attributes as Html exposing (css, maxlength, pattern, placeholder, tabindex, value)
+import Html.Styled.Attributes as Html
+    exposing
+        ( css
+        , maxlength
+        , pattern
+        , placeholder
+        , tabindex
+        , value
+        )
 import Html.Styled.Events exposing (keyCode, on, onBlur, onClick, onInput)
 import Json.Decode as Json
 import Maybe.Extra as Maybe
--- a/src/Nordea/Components/TextInput.elm
+++ b/src/Nordea/Components/TextInput.elm
@@ -36,13 +36,14 @@ import Css
         , opacity
         , outline
         , padding2
-        , padding4
         , paddingLeft
         , paddingRight
         , pct
         , pointer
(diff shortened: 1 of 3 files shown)


src/Nordea/Components/TextInput.elm
```diff
@@ -184,77 +178,41 @@ onEnterPress msg =
 
 view : List (Attribute msg) -> TextInput msg -> Html msg
 view attributes (TextInput config) =
-    if Maybe.isJust config.onClearInput then
```
fredrikloberg, line 187: Hmm, think I changed my mind a little 🙈 What do you think about keeping the setup as it was, but keep the changes you made to the input to this component? (change from `Maybe msg` to `Bool`)

Instead of having `, config.onClearInput |> Maybe.map onClick` as the onClick to the button, we can have `, config.onInput |> Maybe.map (\onInput -> onClick (onInput ""))`. That way, the consumers of this component does not need to setup a new msg and update logic to use it.

2pacalypse: Yes, webkit thingy is taking things out of the Elm a lot. Setting "" is good

src/Nordea/Components/TextInput.elm
```diff
@@ -184,77 +184,65 @@ onEnterPress msg =
...
+                                , padding4 (rem 0.35) (rem 0.25) (rem 0.25) (rem 0.25)
+                                , right (rem 0)
+                                , cursor pointer
+                                ]
+                            )
+                        , config.onInput |> Maybe.map (\onInput -> onClick (onInput ""))
+                        , Just (tabindex 0)
+                        ]
+                    )
+                    |> showIf (config.value |> String.isEmpty |> not)
```
fredrikloberg, line 231: Instead of the if else
```suggestion
                    |> showIf config.hasClearIcon
```

fredrikloberg (approved):
Great!

PR #328, opened 2023-10-26
remove pattern from textinput

2pacalypse:
seems non functional

image

### DIFF
--- a/explorer/src/Stories/TextInput.elm
+++ b/explorer/src/Stories/TextInput.elm
@@ -38,14 +38,6 @@ stories =
                     |> TextInput.view []
           , {}
           )
-        , ( "Pattern only digits"
-          , \_ ->
-                TextInput.init ""
-                    |> TextInput.withPlaceholder "Type something"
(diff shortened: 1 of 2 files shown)


fredrikloberg:
Hmm, I think it was working at some point. Based on [this](https://stackoverflow.com/questions/47769289/html-pattern-to-only-accept-numbers) also, it looks like it should work 🤔

2pacalypse:
> Hmm, I think it was working at some point. Based on [this](https://stackoverflow.com/questions/47769289/html-pattern-to-only-accept-numbers) also, it looks like it should work 🤔

I checked it a bit and you are right. This should work but the trick is it is gonna probably work only when we submit the form as you can check https://www.w3schools.com/tags/tryit.asp?filename=tryhtml5_input_pattern

There seems to be variations that make use of `:valid` and `:invalid` selectors.

https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/pattern

but anyway, best not to remove it.

PR #331, opened 2023-11-08
PAR-2139: support partially filled status labels

2pacalypse:
image

### DIFF
--- a/explorer/src/Stories/Status.elm
+++ b/explorer/src/Stories/Status.elm
@@ -31,7 +31,7 @@ stories =
           , {}
           )
         , ( "Mixed"
-          , \_ -> Status.twoColors "Control prerequisites" Color.yellowStatus (Color.yellow |> Color.withAlpha 0.4) 60.0 []
+          , \_ -> Status.partiallyFilled "Control prerequisites" Color.yellowStatus 60.0 []
           , {}
           )
         ]
--- a/explorer/src/Stories/Status.elm
+++ b/explorer/src/Stories/Status.elm
@@ -1,6 +1,7 @@
 module Stories.Status exposing (..)
 
 import Nordea.Components.Status as Status
+import Nordea.Resources.Colors as Color
 import UIExplorer exposing (Config, UI)
 import UIExplorer.Styled exposing (styledStoriesOf)
 
@@ -27,6 +28,10 @@ stories =
           )
         , ( "Blue"
           , \_ -> Status.blue "Status" []
+          , {}
+          )
+        , ( "Mixed"
+          , \_ -> Status.twoColors "Control prerequisites" Color.yellowStatus (Color.yellow |> Color.withAlpha 0.4) 60.0 []
           , {}
           )
         ]
--- a/src/Nordea/Components/Status.elm
+++ b/src/Nordea/Components/Status.elm
@@ -14,7 +14,6 @@ import Css
         , height
         , hidden
         , inlineBlock
-        , int
         , left
         , overflow
         , padding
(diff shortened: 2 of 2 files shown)


explorer/src/Stories/Status.elm
```diff
@@ -29,4 +30,8 @@ stories =
           , \_ -> Status.blue "Status" []
           , {}
           )
+        , ( "Mixed"
+          , \_ -> Status.twoColors "Control prerequisites" Color.yellowStatus (Color.yellow |> Color.withAlpha 0.4) 60.0 []
```
fredrikloberg, line 34: Could we use the `progress` html tag for this case? Then I think screen readers can pick up the semantic meaning

2pacalypse: wasn't aware of that tag, nice!

src/Nordea/Components/Status.elm
```diff
@@ -85,3 +132,8 @@ blue text =
 gray : String -> List (Attribute msg) -> Html msg
 gray text =
     view text LightGray
+
+
+twoColors : String -> Color -> Color -> Float -> List (Attribute msg) -> Html msg
+twoColors text colorDone colorUndone pctDone =
```
fredrikloberg, line 138: What do you think about making the `alpha 0.4` of the color an internal detail?
```suggestion
twoColors : String -> Color -> Float -> List (Attribute msg) -> Html msg
twoColors text colorDone pctDone =
```

explorer/src/Stories/Status.elm
2pacalypse: I see this tag is problematic in some ways. Firefox does not have support for pseudo elements like "before" and I see using custom html/css for this is recommended. 
https://stackoverflow.com/questions/41429906/how-to-display-data-label-inside-html5-progress-bar-cross-browser/41430146#41430146
https://stackoverflow.com/questions/53278520/how-do-i-put-text-into-a-progress-bar-in-html5-with-bulma

2pacalypse: It is good on chrome and can be worked out 
image

but on firefox it is pretty bad without the pseudo elements
image

fredrikloberg: Ahh ok. Should make sure we have sufficient meta data for accessibility then. Maybe the data-label they use in that thread is enough. There is also a progressbar role: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/progressbar_role

2pacalypse: I see the concern around accessibility. I have made modifications and for example when I click on a row of this table, it reads the columns one by one and for this one it says "Kontroller forutsetninger 50 percent". I think it is good.
image

src/Nordea/Components/Status.elm
2pacalypse: Did we btw abandon the convention of stacking calls like `withTitle` `withProgressBar` with these components? Here `withPartial Float` makes sense on top of already inited bar with color.

explorer/src/Stories/Status.elm
fredrikloberg: Nice!

src/Nordea/Components/Status.elm
fredrikloberg: I think the premiss of the `withPartiallyFilled Float` is a great option indeed! But I am a bit more fan (i think :P)  of having such params as a list instead of the builder pattern, e.g., like Coachmark

PR #333, opened 2023-11-16
PAR-2720: Add small size for text and number input

anisha-deverashetty:
Add small size component so we can use in FF
https://www.figma.com/file/d7rS5EkSr1xMLupCrf0VTh/FinansFront-Sketches?type=design&node-id=90-5800&mode=design&t=pPkrv2KY7Nl6FUnk-0

image

### DIFF
--- a/src/Nordea/Components/NumberInput.elm
+++ b/src/Nordea/Components/NumberInput.elm
@@ -9,6 +9,7 @@ module Nordea.Components.NumberInput exposing
     , withOnBlur
     , withOnInput
     , withPlaceholder
+    , withSmallSize
     , withStep
     )
 
@@ -35,7 +36,13 @@ type alias Config msg =
     , showError : Bool
     , onBlur : Maybe msg
     , formatter : Maybe (Float -> String)
+    , size : Size
     }
+
+
+type Size
+    = Small
+    | Standard
 
 
 type NumberInput msg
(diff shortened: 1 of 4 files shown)


src/Nordea/Components/NumberInput.elm
```diff
@@ -152,14 +165,25 @@ getStyles config =
 
             else
                 Colors.mediumGray
+
+        sizeSpecificStyling =
+            case config.size of
+                Small ->
+                    [ fontSize (rem 0.75)
+                    , height (rem 1.5)
+                    , padding2 (rem 0.25) (rem 0.5)
```
hannabergljot, line 174: Is it correct that it has more padding than the standard size?☺️

hannabergljot (approved):
Really nice!!✨

src/Nordea/Components/NumberInput.elm
anisha-deverashetty: I think this is correct. Has less padding on the left and right `0.5` then the standard `0.75`. The padding on top/bottom doesn't make much difference here. I had the values from design so have included it. We can also set it `rem 0`

hannabergljot: I see, if Lines design has 0.25 then I think we should stick with that, was just curious😁

anisha-deverashetty: Good comment though. So, then lets keep the top/bottom padding ☺️

PR #335, opened 2023-11-17
Refactor small size button component

hannabergljot:
Refactored the small versions of the button component to align with the new design style for FinansFront. Also rewrote to match the same `withSmallSize` patters as the other sized components.

Before:
Screenshot 2023-11-17 at 12 54 25 PM

After:
Screenshot 2023-11-17 at 12 54 48 PM


Rostislav013 (approved):

fredrikloberg (approved):
🙌

PR #336, opened 2023-11-17
Add small size dropdown component

hannabergljot:
Added a small size version of the dropdown component, to align with design decisions made for FinansFront.

Screenshot: 
Screenshot 2023-11-27 at 1 43 41 PM

### DIFF
--- a/src/Nordea/Resources/Icons.elm
+++ b/src/Nordea/Resources/Icons.elm
@@ -12,6 +12,7 @@ module Nordea.Resources.Icons exposing
     , chevronDown
     , chevronDownBolded
     , chevronDownFilled
+    , chevronDownFilledSmall
     , chevronLeft
     , chevronRight
     , chevronUp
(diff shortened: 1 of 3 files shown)


src/Nordea/Resources/Icons.elm
```diff
@@ -226,6 +211,34 @@ chevronDownFilled attrs =
         ]
 
 
+chevronDownFilledSmall : List (Attribute msg) -> Html msg
+chevronDownFilledSmall attrs =
```
fredrikloberg, line 215: Is this a different icon, and not just a smaller version of the same icon?

hannabergljot: It is a smaller version of the normal one, but just shrinking the original doesn't work, because the ratio between background and the arrow itself is different. So to keep the size of the arrow, while reducing the size of the whole box, we needed to create a new icon☺️ Could be a method to do this that I'm not aware of though? Just scaling it down at least did not work

PR #338, opened 2023-11-23
PAR-2746: Add small label component

anisha-deverashetty:
Add small version of label which could be used with the small version of other components

image
image
image


fredrikloberg (approved):
🙌

hannabergljot (approved):
Perfect!!✨

PR #340, opened 2023-11-23
PAR-2745: button with flatlink style does not show focus

2pacalypse:
The design library has the focus
https://www.figma.com/file/tgFkTfmNZE0zwx44OeaohE/New-Design-System-anno-2023?type=design&node-id=399-25727&mode=design&t=iPshDl7nyMSjZ8HB-0

image

There are alternative ways of doing this

- outline
  - only text clickable at any point
  - the outline might overlay other elements

outline

- padding
  - makes the empty space clickable along with text

padding

- padding only on focus
  - the space is only clickable in the focus state 
  - it adds the padding only on focus so causes some movement in the layout   

paddingonfocus

### DIFF
--- a/src/Nordea/Components/Button.elm
+++ b/src/Nordea/Components/Button.elm
@@ -15,67 +15,7 @@ module Nordea.Components.Button exposing
     , withStyles
     )
 
-import Css
-    exposing
-        ( Style
-        , absolute
-        , alignItems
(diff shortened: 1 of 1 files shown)


fredrikloberg (approved):
💯

Rostislav013 (approved):

PR #342, opened 2023-11-24
PAR-2742: new arrow placements for tooltip

2pacalypse:
The design library has more placements than just the arrow being on the center.

https://www.figma.com/file/tgFkTfmNZE0zwx44OeaohE/New-Design-System-anno-2023?type=design&node-id=8-1997&mode=design&t=2juhpiu62fIbnojD-0

image

Introduce 3 variants, `NearCorner, NearOtherCorner, Center` for arrow placement.
image

### DIFF
--- a/explorer/src/Stories/Tooltip.elm
+++ b/explorer/src/Stories/Tooltip.elm
@@ -41,7 +41,7 @@ stories =
                         |> Tooltip.withPlacement Top
                         |> Tooltip.withVisibility Tooltip.Show
                         |> Tooltip.withContent (Tooltip.infoTooltip [] [ text "This is a tooltip" ])
-                        |> Tooltip.view [] [ text "here" ]
+                        |> Tooltip.view [] [ text "here here here here here here here here here" ]
                     ]
           , {}
           )
@@ -53,7 +53,7 @@ stories =
                         |> Tooltip.withPlacement TopRight
                         |> Tooltip.withVisibility Tooltip.Show
                         |> Tooltip.withContent (Tooltip.infoTooltip [] [ text "This is a tooltip" ])
-                        |> Tooltip.view [] [ text "here" ]
+                        |> Tooltip.view [] [ text "here here here here here here here here here" ]
                     ]
           , {}
           )
@@ -65,7 +65,7 @@ stories =
                         |> Tooltip.withPlacement TopLeft
                         |> Tooltip.withVisibility Tooltip.Show
                         |> Tooltip.withContent (Tooltip.infoTooltip [] [ text "This is a tooltip" ])
-                        |> Tooltip.view [] [ text "here" ]
+                        |> Tooltip.view [] [ text "here here here here here here here here here" ]
                     ]
           , {}
           )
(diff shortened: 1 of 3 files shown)


explorer/src/Stories/Tooltip.elm
```diff
@@ -40,6 +40,30 @@ stories =
...
+                    ]
+          , {}
+          )
+        , ( "Top with arrow placed near other corner"
+          , \_ ->
+                div []
+                    [ text "There is a tooltip "
+                    , Tooltip.init
+                        |> Tooltip.withPlacement Top
+                        |> Tooltip.withArrowPlacement NearOtherCorner
```
fredrikloberg, line 61: Instead of introducing another field, what do you think about extending the existing? One benefit is that we get more clear naming, instead of `Top` and `NearCorner`, `Center`, `NearOtherCorner`, we can for example have `TopLeft`, `Top` (or TopCenter), and `TopRight`. (And for example `RightBottom` for bottom corner on the right side)

What do you think?
```suggestion
                        |> Tooltip.withPlacement TopLeft
```

fredrikloberg: Css already have similar naming scheme like this suggestion for corners: see https://package.elm-lang.org/packages/rtfeldman/elm-css/latest/Css#borderBottomRightRadius

src/Nordea/Components/Tooltip.elm
```diff
@@ -123,46 +136,123 @@ withVisibility visibility (Tooltip config) =
 view : List (Attribute msg) -> List (Html msg) -> Tooltip msg -> Html msg
 view attrs children (Tooltip config) =
     let
+        arrowDistanceFromCornerInPct =
+            15
```
fredrikloberg, line 140: Is the percentage relative to the width/height of the tooltip? If that is the case is it better to make it static?

fredrikloberg (approved):
🙌

src/Nordea/Components/Tooltip.elm
2pacalypse: Yes, 15% is calculated depending on the width/height of the content. The way it works right now is arrow stays central always to the content being tooltipped, but the positioning of the box changes.
By static, if you mean dropping the pct and doing it with rem, It might look a bit odd for wide or tall boxes. It is more robust to height/width changes this way I think.

fredrikloberg: The need for these extra arrow placements is to be able to keep the tooltip on-screen. So my concern was whether the relative positioning may push parts of the tooltip outside the screen as the size increases. For example for our coachmark case, double the width and I guess the right side may be outside the screen? There may be a higher risk for mobile views where we have much less padding and the icon appears closer to the edge.

Not sure if a realistic concern tho, have not tested 😄

2pacalypse: you are right it might break in that way and go out of screen for large input. might make sense to make it static.

It is especially tricky to make the horizontal one work because the height is so short there. I will have another look if it generalizes good.
image

PR #343, opened 2023-11-28
PAR-2747: Adjustments to small text input to use with icons

anisha-deverashetty:
Fix padding on Standard input field 
image
image
Small with search icon
image
Small with currency
image
Small with cancel icon
image

### DIFF
--- a/explorer/src/Stories/TextInput.elm
+++ b/explorer/src/Stories/TextInput.elm
@@ -82,9 +82,17 @@ stories =
                     |> TextInput.view []
           , {}
           )
-        , ( "With small size"
+        , ( "Small"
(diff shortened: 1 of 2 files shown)


fredrikloberg (approved):
Looks good! Is it not relevant to have search icon for the small size?

fredrikloberg:
> Also not allow search and cancel icons with small input field since the icons are quite big. We can add smaller icons later if needed.

I didnt read the description first 😅 But the size of the icon can be adjusted with the width/height css property to fit the small size. Does this look weird?

fredrikloberg (approved):
Nice!

PR #344, opened 2023-11-30
Add small modal component

anisha-deverashetty:
Small version of modal component which will be used in FF.
image

### DIFF
--- a/src/Nordea/Components/Modal.elm
+++ b/src/Nordea/Components/Modal.elm
@@ -1,6 +1,7 @@
 module Nordea.Components.Modal exposing
     ( default
     , newsModal
+    , small
     , view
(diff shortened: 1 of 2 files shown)


src/Nordea/Components/Modal.elm
```diff
@@ -58,6 +60,11 @@ newsModal msg =
     init msg NewsModal
 
 
+small : msg -> Modal msg
```
hannabergljot, line 63: Could we use withSmallSize so that it's more consistent with other components here?☺️

anisha-deverashetty: Yes, I think that's a good idea. Will update

PR #348, opened 2023-12-19
Add table that is shared between PH and KP

atraastrum:
Since this component is shared between two teams. It makes sense to add it to component library.

Will add example with sortable columns later.


hhaga (approved):
💯

wenwei-lai (approved):
LGTM :rocket:

PR #350, opened 2023-12-22
Make it possible to have right aligned column

atraastrum:
Allow to style text inside table cells to make them right aligned in a backward compatible way.

### DIFF
--- a/src/Nordea/Components/SortableTable.elm
+++ b/src/Nordea/Components/SortableTable.elm
@@ -8,6 +8,7 @@ module Nordea.Components.SortableTable exposing
     , mobileElementFull
     , mobileTextElement
     , sortableTextHeader
+    , sortableTextHeaderStyled
     , sortingAsDict
     , sortingToString
     , tbody
@@ -72,11 +73,22 @@ headerElement attrs =
         (css [ flex (num 1), pseudoClass "not(:first-child)" [ marginLeft (rem 1) ] ] :: attrs)
 
 
-sortableTextHeader : { css : List Css.Style, sorting : Maybe Sorting, onClick : msg, label : String } -> Html msg
-sortableTextHeader config =
+sortableTextHeaderStyled : { css : List Css.Style, textCss : List Css.Style, chevronAlignRight : Bool, sorting : Maybe Sorting, onClick : msg, label : String } -> Html msg
+sortableTextHeaderStyled config =
     let
         chevronStyle =
-            [ css [ flexShrink (int 0), width (rem 0.5), marginLeft (rem 1) ] ]
+            [ css
+                [ flexShrink (int 0)
+                , width (rem 0.5)
+                , (if config.chevronAlignRight then
+                    marginRight
+
+                   else
+                    marginLeft
+                  )
+                    (rem 1)
+                ]
+            ]
 
         scrollDirectionChevron =
             case config.sorting of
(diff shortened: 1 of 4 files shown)


atraastrum:
[Align column with amount to the right on the customer invoices page](https://trello.com/c/g6IUiUoj/1314-align-column-with-amount-to-the-right-on-the-customer-invoices-page)

fredrikloberg:
Do you have a screenshot of where right aligned is needed?

src/Nordea/Components/SortableTable.elm
```diff
@@ -110,38 +128,55 @@ sortableTextHeader config =
...
 
-textHeader : { css : List Css.Style, label : String } -> Html msg
-textHeader config =
+sortableTextHeader : { css : List Css.Style, sorting : Maybe Sorting, onClick : msg, label : String } -> Html msg
+sortableTextHeader config =
+    sortableTextHeaderStyled { css = config.css, textCss = [], chevronAlignRight = False, sorting = config.sorting, onClick = config.onClick, label = config.label }
+
+
+textHeaderStyled : { css : List Css.Style, textCss : List Css.Style, label : String } -> Html msg
+textHeaderStyled config =
```
fredrikloberg, line 154: Taking in multiple css params makes the abstraction leaky, as it exposes the internal html structure of the component a bit. Can we avoid it by: 
```elm
textHeader : { css : List Css.Style, label : String } -> Html msg
textHeader config =
    Text.textTinyLight
        |> Text.withHtmlTag headerElement
        |> Text.view
            [ Attrs.title config.label
            , css
                ([ displayFlex
                 , textOverflow ellipsis
                 , overflow hidden
                 , whiteSpace noWrap
                 , color Color.eclipse
                 , css (flex (num 1)
                 ]
                    ++ config.css
                )
            ]
            [ Html.text config.label ]
```

(Same for `sortableTextHeaderStyled` above.) 

This way, we also keep the table interface simples with fewer options.

atraastrum:
> Do you have a screenshot of where right aligned is needed?

Find the screen shot where right aligned columns are used.
image


How it used to be with custom table
image

This change was requested by Andreas.

fredrikloberg:
> Find the screen shot where right aligned columns are used. 
> How it used to be with custom table 
> This change was requested by Andreas.

Thanks! So do I understand it correctly that numbers/currency only should be right aligned, and that this should be a part of our design system? If so, does it make sense to make it more explicit by creating `numberHeader` or `currencyHeader` or something? This way, it is slightly harder to forget, especially across teams and projects.

atraastrum:
> > Find the screen shot where right aligned columns are used.
> > How it used to be with custom table
> > This change was requested by Andreas.
> 
> Thanks! So do I understand it correctly that numbers/currency only should be right aligned, and that this should be a part of our design system? If so, does it make sense to make it more explicit by creating `numberHeader` or `currencyHeader` or something? This way, it is slightly harder to forget, especially across teams and projects.

I like the idea. But not sure should it be part of design. At least on the internet some people advice that numeric data is right aligned

src/Nordea/Components/SortableTable.elm
atraastrum: Introduce more explicit APIs here 16da6915a2d1c5f60b262231894eb636d3b05769

Do you think the change you suggested is still needed/ Now internal details are not exposed anymore?

fredrikloberg: Not strictly needed, but I think it is still a nice improvement

fredrikloberg (approved):

src/Nordea/Components/SortableTable.elm
atraastrum: I will do it in separate PR

PR #353, opened 2023-12-29
fix tooltip width and z-index issue

2pacalypse:
`min-width` was removed some time ago from tooltip and later it caused issues for `DropdownFilter`, and as part of fix it was added again here https://github.com/SGFinansAS/elm-components/pull/345

But the fix there also caused an issue in tooltip like below

image

The tooltip lost its position at the center, and `min-width` caused the tooltip to cover a very large area (as big as the element that is being tooltipped). This is bad because you cannot click anything because of z-index.

Now with this fix, both dropdown and tooltip should work as intended.

image
image


fredrikloberg (approved):
Nice!

PR #354, opened 2024-01-05
add snackbar component

fonnes:
Copied from partnerhub because we want to use it in the customer portal

image


fredrikloberg (approved):
🙌

PR #359, opened 2024-02-08
Add gap css function

morteako:
Annoying and error prone to write Css.property "gap" "1rem" instead of `gap (rem 1)` like you would for height,length and such. 
This fixes that :))
Also easier to read and more in line with most other props
Copied the type from the Css-library, so it works with rem/px/pct and such

* gap (rem 1)
 image
* gap (px 1)
image


fredrikloberg (approved):
🙌

fonnes (approved):
Love it! 🚀🚀🚀

PR #363, opened 2024-03-01
Add helper functions for CSS properties related to grids

kjesvale:
Add helper functions for properties that are related to CSS Grids.

Elm-css is a bit outdated and does not support newer CSS syntax. This is a continuation of @morteako 's idea of adding the missing functions to the component library (`gap` and `gap2` in his case).

I haven't done the effort of typing these functions in a sophisticated way, so most just accept a single _string_ parameter. Adding better types would be a greater effort since the Grid spec introduces a bunch of new units that are also not present in elm-css. Anyway, I think it adds some value to be able to type:

```elm
css
  [ displayGrid
  , gridTemplateColumns "repeat(4, 1fr)"
  ]
```

Instead of:

```elm
css
  [ Css.property "display" "grid"
  , Css.property "grid-template-columns" "repeat(4, 1fr)"
  ]
```

Added some links to MDN as well to get some help right in the editor.

### DIFF
--- a/src/Nordea/Css.elm
+++ b/src/Nordea/Css.elm
@@ -156,8 +156,8 @@ gridTemplateColumns columns =
 {-| Sets [`grid-template-rows`](https://developer.mozilla.org/en-US/docs/Web/CSS/grid-template-rows).
 -}
 gridTemplateRows : String -> Css.Style
-gridTemplateRows columns =
-    Css.property "grid-template-rows" columns
+gridTemplateRows rows =
+    Css.property "grid-template-rows" rows
 
 
 {-| Sets [`gap`](https://developer.mozilla.org/en-US/docs/Web/CSS/gap).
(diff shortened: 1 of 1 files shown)


src/Nordea/Css.elm
```diff
@@ -13,16 +34,181 @@ module Nordea.Css exposing
...
+justifySelf alignment =
+    Css.property "justifySelf" alignment
+
+
+{-| Sets [`justify-items`](https://developer.mozilla.org/en-US/docs/Web/CSS/justify-items).
+-}
+justifyItems : String -> Css.Style
+justifyItems alignment =
+    Css.property "justifyItems" alignment
+
```
fredrikloberg, line 210: Do these support the same values as alignSelf, justifyContents etc? if so, use `(ExplicitLength IncompatibleUnits -> Style)` as the param type?

fredrikloberg (approved):
🙌

PR #369, opened 2024-03-26
DropdownFilter: Add Placeholder and withHasError-function

morteako:
# Expand DropdownFilter
Adds the option to add a placeholder (`placeholder : Maybe String` and `withPlaceholder`) 

Adds `withHasError` function where you can specifically set the error, instead of it only coming from 0 matches.
The model already has `hasError : Bool` , but it is defaulted to False and previously not possible to change.

image

image


2pacalypse (approved):
nice additions to catch up with other components :100:

PR #370, opened 2024-03-27
Add pagination component

2pacalypse:
We have the need deriving from this [sketch](https://www.figma.com/file/d7rS5EkSr1xMLupCrf0VTh/FinansFront-Sketches?type=design&node-id=1474-4018&mode=design&t=ifN97oa8pyef67Wf-4)

___
From Elm view
image


src/Nordea/Components/Pagination.elm
```diff
@@ -0,0 +1,181 @@
...
+                        visibility hidden
+
+                      else
+                        visibility visible
+                    ]
+                ]
+                ( properties.currentPage - 1, "Previous" )
+                |> showIf properties.showPrevButton
+
+        next =
```
morteako, line 79: nextButton

src/Nordea/Components/Pagination.elm
```diff
@@ -0,0 +1,181 @@
...
+
+            else
+                l
+
+        adjustTailOffset l =
+            if List.member properties.totalPages l then
+                l |> List.map (\n -> n - 1)
+
+            else
+                l
```
morteako, line 151: should probably rename `l` to something a bit more clearifying

src/Nordea/Components/Pagination.elm
```diff
@@ -0,0 +1,205 @@
...
+                   )
+                ++ [ properties.totalPages ]
+                |> List.unique
+
+        addDots ns =
+            case ns of
+                a :: b :: c ->
+                    let
+                        ( aVal, bVal ) =
+                            ( min (Tuple.first a) (Tuple.first b), max (Tuple.first a) (Tuple.first b) )
```
morteako, line 170: (minVal, maxVal)

src/Nordea/Components/Pagination.elm
```diff
@@ -0,0 +1,205 @@
...
+                             --override the focus and underline from the main component
+                            )
+                        ]
+                        [ Text.init TextTinyHeavy |> Text.view [] [ Html.text textShownForPage ] ]
+                ]
+
+        prev =
+            optionalConfig.prevButton
+                |> Maybe.map
+                    (\s ->
```
morteako, line 85: s -> prevButton or something

morteako:
Seems to look good to me :))

fredrikloberg (commented):
Unrelated to the implementation, but have you double checked if there is actually a need for it? We do it differently in PH with a much simpler "show more" feature. I feel like it may not be needed and the much simpler solution we have in PH might actually solve the problem just as well. My assumption here is that a user dont want to go to a specific page when looking for cases, they either narrow the search or occasionally hit "show more" a few times.)

What do you think?

2pacalypse:
> Unrelated to the implementation, but have you double checked if there is actually a need for it? We do it differently in PH with a much simpler "show more" feature. I feel like it may not be needed and the much simpler solution we have in PH might actually solve the problem just as well. My assumption here is that a user dont want to go to a specific page when looking for cases, they either narrow the search or occasionally hit "show more" a few times.)
> 
> What do you think?

There seems to be that convention yes, both in KP and PH. If I recall good, the preference to new component could be due to the section being smaller than what we usually have in PH, KP. There we have the whole page, but here we have a scroll and we fetch 50 applications per page.  When we load more and more there exists a lot in a tiny section. I am not sure though if this was what drove Line to have a new component like this. Personally, I like pages more than infinite loading. They remind me of gmail when I want to delete/move everything, and also let us estimate the total and imo they look better. This is not a strong reason to use it though. Also, chances are low as you point out but showing 3k rows crashed for me in FF, and halted for a few seconds in Chrome.

Both 1 2 3, and show more have that same problem you mentioned. If they are looking for a specific contract, they will have to go on a journey either way because we do not have a search bar. I considered I could order by ident and at least let them improvise something like binary search with dots to jump and CTRL+F, then this pagination is better slightly in that way than loading sequentially.

I will ask Line, I am not sure if she has plans to use this in other places maybe she sets up the ground for other things.

fredrikloberg:
Yes I agree with the reasoning. My only concern is that I have a doubt that it will be used. If it is used in 1% of cases I dont consider it worth it. And even if it is used, I feel like that is a symptom of the lack of filtering/searching

fredrikloberg (approved):
code looks good!

LineHjertas:
I agree with what the both of you are saying.
Personally I do not like pagination, very rarely do I go through page after page looking for something without getting incredibly frustrated. But it does also feel very overwhelming to click show more many many times and we do need to give them the chance to see all the contracts.
The user need when clicking through the pages is that I am looking for something and I cant filter down? What has a higher cost to implement pagination or to add a search or more filter? I am hoping to not have to use pagination more then necessary as it is not really an answer to the user need. In this case I would assume its a cost question..

LineHjertas:
I have done zero research into this so I do not know the user need in that sense. But I am assuming that you would have one or a few contracts in mind that you want to filter on.

PR #372, opened 2024-04-19
Black magic: Add "Onclick outside" support

fredrikloberg:
Reacting when you click outside of an element is a big pain. For simple components, simply using the `focusout` event is the way to go. The problem arises when you have interactive sub components inside. Take the multi select dropdown for example. You click it to open it, and when you click outside it closes (with the focusout event). But, when you interact with a component inside it, the focusout event is still fired because the focus moves to a sub component. Trying to make it behave correctly is very challenging and is timing sensitive because the browser is free to fire the events in any order it wants.

The only proper solution I have found is through JS. This PR fixes the issue by inlining some JS code through an iframe element.


fonnes (approved):
LGTM!

PR #377, opened 2024-05-30
Add date picker component

anisha-deverashetty:
Add date picker component 

image

https://www.figma.com/design/tgFkTfmNZE0zwx44OeaohE/New-Design-System-anno-2023?node-id=16-16701&t=cX7W8hkrEqrIQiaj-0

### DIFF
--- a/src/Nordea/Components/DatePicker.elm
+++ b/src/Nordea/Components/DatePicker.elm
@@ -1,4 +1,4 @@
-module Nordea.Components.DatePicker exposing (OptionalConfig(..), view)
+module Nordea.Components.DatePicker exposing (InternalState(..), OptionalConfig(..), init, view)
 
 import Css
     exposing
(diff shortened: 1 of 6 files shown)


src/Nordea/Components/DatePicker.elm
```diff
@@ -0,0 +1,374 @@
...
+import Time exposing (Month, Weekday)
+
+
+type alias Config msg =
+    { hasFocus : Bool
+    , onFocus : Bool -> msg
+    , input : String
+    , onInput : String -> msg
+    , today : Date
+    , onSelect : String -> msg
```
fredrikloberg, line 71: Could this return a Date directly?

anisha-deverashetty: Yes, we can return date and remove the formatter.

src/Nordea/Components/DatePicker.elm
```diff
@@ -0,0 +1,374 @@
...
+    , onMonthChange : Month -> Int -> msg
+    , selectedMonth : Maybe ( Month, Int )
```
fredrikloberg, line 73: Could it be beneficial to hide some of these fields a little to make the component easier to use? For example create an opaque type `InternalState`:

```elm
type InternalState =
    { hasFocus : Bool
    , onFocus : Bool -> msg
    , input : String
    , onInput : String -> msg
    , today : Date
    , onMonthChange : Month -> Int -> msg
    , selectedMonth : Maybe ( Month, Int )
    }

type alias Config msg =
    { onSelect : Date -> msg
    , currentValue: Maybe Date
    , internalState : InternalState
    , onInternalStateChange : InternalState -> msg
    }

init onSelect currentValue today onSelect onInternalStateChange =
    {construct config with internal state}
``` 

This way, the user only has to be concerned with the a few fields and the component becomes free to change the internal state without needing to update any of the uses.

What do you think?

fredrikloberg (approved):
Great work! 🙌

src/Nordea/Components/DatePicker.elm
anisha-deverashetty: I have updated it based on your comment 👍

src/Nordea/Components/DatePicker.elm
```diff
@@ -0,0 +1,410 @@
+module Nordea.Components.DatePicker exposing (InternalState(..), OptionalConfig(..), init, view)
```
fredrikloberg, line 1: Could we make the internal state opaque? That way, if we want to make changes to it, we only need to update this file. The only way to construct it is via functions in this file, for example the init function
```suggestion
module Nordea.Components.DatePicker exposing (InternalState, OptionalConfig(..), init, view)
```

PR #382, opened 2024-06-06
Date picker improvements

anisha-deverashetty:
Small improvements:

- When we use multiple same instances of date picker in a page, then clicking anywhere inside will close it. So needs stop propagation on click.

- Clicking prev/next month as soon as opening datepicker, sends the showError msg. Fix it so that showError is send only during onBlur when the calendar is closed or after choosing a date.

### DIFF
--- a/src/Nordea/Components/DatePicker.elm
+++ b/src/Nordea/Components/DatePicker.elm
@@ -141,7 +141,7 @@ view attrs optional (DatePicker config) =
         internalState =
             internalStateValues config.internalState
 
-        { placeholder, parser, formatter, firstDayOfWeek, onBlur, error } =
+        { placeholder, parser, formatter, firstDayOfWeek, showError, error } =
             optional
                 |> List.foldl
                     (\e acc ->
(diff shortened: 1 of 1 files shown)


fredrikloberg (commented):
Why is onBlur needed/what is it used for?

anisha-deverashetty:
> Why is onBlur needed/what is it used for?

It is used to set the `error` to true/false. I think better to rename it to showError or something.

fredrikloberg (approved):
Could you explain the point with multiple instances further? I dont quite see why having multiple makes it close when clicking inside 🤔

anisha-deverashetty:
> Could you explain the point with multiple instances further? I dont quite see why having multiple makes it close when clicking inside 🤔
Having multiple is fine but have multiple same instance is causing issue. For example in the Inventory asset details page we have 3 same datepicker instance for invoice date to support responsive style. We always show 1 and hide the rest. When clicking inside the datepicker, we are getting the "outsideclick" event from hidden ones which will set `onFocus=False` and makes datepicker to close.

PR #383, opened 2024-06-13
Add withId so that we can give an Dom id to the input element

rogerab79:
- This allows us to do things like direct focus on the input element (using Browser.Dom).

### DIFF
--- a/src/Nordea/Components/TextInput.elm
+++ b/src/Nordea/Components/TextInput.elm
@@ -185,6 +185,7 @@ withId : String -> TextInput msg -> TextInput msg
 withId id (TextInput config) =
     TextInput { config | id = Just id }
 
+
 onEnterPress : msg -> Attribute msg
 onEnterPress msg =
     let
--- a/src/Nordea/Components/TextInput.elm
+++ b/src/Nordea/Components/TextInput.elm
@@ -5,6 +5,7 @@ module Nordea.Components.TextInput exposing
     , withClearInput
     , withCurrency
     , withError
+    , withId
     , withMaxLength
     , withOnBlur
     , withOnEnterPress
@@ -13,7 +14,6 @@ module Nordea.Components.TextInput exposing
     , withPlaceholder
     , withSearchIcon
     , withSmallSize
-    , withId
     )
 
 import Css
@@ -91,7 +91,7 @@ type alias Config msg =
     , onEnterPress : Maybe msg
     , currency : Maybe String
     , size : Size
-    , id: Maybe String
+    , id : Maybe String
     }
 
 
--- a/src/Nordea/Components/TextInput.elm
+++ b/src/Nordea/Components/TextInput.elm
@@ -13,6 +13,7 @@ module Nordea.Components.TextInput exposing
     , withPlaceholder
     , withSearchIcon
     , withSmallSize
+    , withId
     )
 
 import Css
@@ -90,6 +91,7 @@ type alias Config msg =
     , onEnterPress : Maybe msg
     , currency : Maybe String
     , size : Size
+    , id: Maybe String
     }
 
 
@@ -117,6 +119,7 @@ init value =
         , onEnterPress = Nothing
         , currency = Nothing
         , size = Standard
+        , id = Nothing
         }
 
 
@@ -177,6 +180,10 @@ withSmallSize : TextInput msg -> TextInput msg
 withSmallSize (TextInput config) =
     TextInput { config | size = Small }
 
+
+withId : String -> TextInput msg -> TextInput msg
+withId id (TextInput config) =
+    TextInput { config | id = Just id }
 
 onEnterPress : msg -> Attribute msg
 onEnterPress msg =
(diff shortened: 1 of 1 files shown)


fredrikloberg:
You can already set an id attribute by:

```elm
TextInput.init "Text"
    |> TextInput.view [ id "some-id" ]
```

So I would suggest to do that instead

rogerab79:
> You can already set an id attribute by:
> 
> ```elm
> TextInput.init "Text"
>     |> TextInput.view [ id "some-id" ]
> ```
> 
> So I would suggest to do that instead

This does not add the attribute to the input element - which is needed if I want to direct focus to that element.

fredrikloberg:
> This does not add the attribute to the input element - which is needed if I want to direct focus to that element.
> 
> EDIT: Or to be specific, it also adds it to the embedding element - id:s should be unique so must only be added to the input.

Ohh yes, you are correct. Unfortunate complexity of having composite components 🫤 

But what is the actual use case of needing manual focus management? Could [autofocus](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/autofocus) be an option?

rogerab79:
> > This does not add the attribute to the input element - which is needed if I want to direct focus to that element.
> > EDIT: Or to be specific, it also adds it to the embedding element - id:s should be unique so must only be added to the input.
> 
> Ohh yes, you are correct. Unfortunate complexity of having composite components 🫤
> 
> But what is the actual use case of needing manual focus management? Could [autofocus](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/autofocus) be an option?

The use case is in a search inut + dropdown for selecting items. It's desirable to keep the kbd focus on the field even after selecting items (=> Dom.focus) and it makes it easier to manage automatic closing of the dropdown - i.e. onBlur on the search field. Autofocus is obviously not an option here..

fredrikloberg:
> The use case is in a search inut + dropdown for selecting items. It's desirable to keep the kbd focus on the field even after selecting items (=> Dom.focus) and it makes it easier to manage automatic closing of the dropdown - i.e. onBlur on the search field. Autofocus is obviously not an option here..

I see, that sounds like a reasonable use case indeed. I do however have one concern there, how will it work in terms of accessibility? It must to be possible to use well with a keyboard only, and you need to be able to move focus freely between the elements (with your keyboard) for screen readers to work. And you should also consider the point the from link above about how moving focus around automatically can be bad for accessibility in general. 

Do you think it will be fine?

rogerab79:
> > The use case is in a search inut + dropdown for selecting items. It's desirable to keep the kbd focus on the field even after selecting items (=> Dom.focus) and it makes it easier to manage automatic closing of the dropdown - i.e. onBlur on the search field. Autofocus is obviously not an option here..
> 
> I see, that sounds like a reasonable use case indeed. I do however have one concern there, how will it work in terms of accessibility? It must to be possible to use well with a keyboard only, and you need to be able to move focus freely between the elements (with your keyboard) for screen readers to work. And you should also consider the point the from link above about how moving focus around automatically can be bad for accessibility in general.
> 
> Do you think it will be fine?

I'm not too familiar with screen readers but the focus changes only when selecting an item from the list, so it should work when moving the focus around generally.  In fact it could would perhaps work better as the close button will always be at a predictable place from the current focus (next button).

fredrikloberg (approved):

fredrikloberg:
> I'm not too familiar with screen readers but the focus changes only when selecting an item from the list, so it should work when moving the focus around generally. In fact it could would perhaps work better as the close button will always be at a predictable place from the current focus (next button).

It sounds suboptimal accessibility wise because I dont think the screen reader will announce the selected values, it will instead skip the announcement and jump to the text input and announce the text there instead. The behaviour will probably be better for mouse users tho (but keep in mind accessibility is a hard requirement).
Regarding closing, it should be possible to press Esc to close the dropdown. The focus should still be on the dropdown, and pressing space or Enter should open it again.

rogerab79:
> > I'm not too familiar with screen readers but the focus changes only when selecting an item from the list, so it should work when moving the focus around generally. In fact it could would perhaps work better as the close button will always be at a predictable place from the current focus (next button).
> 
> It sounds suboptimal accessibility wise because I dont think the screen reader will announce the selected values, it will instead skip the announcement and jump to the text input and announce the text there instead. The behaviour will probably be better for mouse users tho (but keep in mind accessibility is a hard requirement). Regarding closing, it should be possible to press Esc to close the dropdown. The focus should still be on the dropdown, and pressing space or Enter should open it again.

I'll keep these things in mind; I don't think we have esc-support in any internal dropdown components at CP - I'll add it at least to the current component under development.

It would sure help a lot to have access to some screen-reader software when developing accessibility-friendly components..

fredrikloberg:
> I'll keep these things in mind; I don't think we have esc-support in any internal dropdown components at CP - I'll add it at least to the current component under development.
> 
> It would sure help a lot to have access to some screen-reader software when developing accessibility-friendly components..

Great! I think esc-support might have been lost. I remember we had it in one of the older components at least.

There is built-in support for screen readers on Mac and Windows (I dont know about linux), which is very useful to run through every once in a while.

PR #384, opened 2024-06-13
Add onEscPress to Html.Events

rogerab79:
- It's useful when building accessible components.
- Needed to e.g. close dropdown on ESC.

Also see discussion at the end of https://github.com/SGFinansAS/elm-components/pull/383


fredrikloberg (approved):
🙌

PR #393, opened 2024-07-12
FiveStarRating component

sarastorbjork:
Component for selecting a rating out of five stars. Based on  https://www.figma.com/file/d7rS5EkSr1xMLupCrf0VTh/FinansFront-Sketches?type=design&node-id=1603-235&mode=design

Usage example : https://github.com/SGFinansAS/FinansFront/pull/8539/files

[Screencast from 2024-07-12 12-46-26.webm](https://github.com/user-attachments/assets/cae2f0a2-d467-46ca-9995-8fc0a1761047)

### DIFF
--- a/src/Nordea/Components/FiveStarRating.elm
+++ b/src/Nordea/Components/FiveStarRating.elm
@@ -1,6 +1,6 @@
 module Nordea.Components.FiveStarRating exposing (FiveStarRating(..), init, view)
 
-import Css exposing (alignItems, backgroundColor, borderRadius, center, color, displayFlex, flexDirection, focus, height, justifyContent, none, outline, padding, rem, row, transparent, width)
+import Css exposing (alignItems, backgroundColor, borderRadius, borderStyle, center, color, displayFlex, flexDirection, focus, height, justifyContent, none, outline, padding, rem, row, transparent, width)
 import Html.Styled as Styled exposing (Attribute)
 import Html.Styled.Attributes exposing (css)
 import Html.Styled.Events exposing (onClick, onMouseEnter)
@@ -15,20 +15,14 @@ type FiveStarRating msg
 
 
 type alias Config msg =
-    { rating : Int
-    , currentHover : Int
-    , onHover : Int -> msg
-    , onClick : Int -> msg
+    { setRating : Int -> msg
     }
 
 
-init : (Int -> msg) -> (Int -> msg) -> FiveStarRating msg
-init onHover onClick =
+init : (Int -> msg) -> FiveStarRating msg
+init setRating =
     FiveStarRating
-        { rating = 0
-        , currentHover = 0
-        , onHover = onHover
-        , onClick = onClick
+        { setRating = setRating
         }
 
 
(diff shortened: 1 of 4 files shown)


fredrikljung93 (approved):
LGTM :star::star::star::star::star:

src/Nordea/Components/FiveStarRating.elm
```diff
@@ -0,0 +1,69 @@
...
+                    , backgroundColor transparent
+                    , borderRadius (rem 0.1)
+                    , padding (rem 0.25)
+                    , focus
+                        [ outline none
+                        , borderStyle none
+                        , Css.property "box-shadow" ("0rem 0rem 0rem 0.2rem " ++ Themes.colorVariable Colors.haasBlue)
+                        ]
+                    ]
+                , onMouseEnter (config.setRating value)
```
2pacalypse, line 63: If I understand correctly, this rating component works purely by hovering, and not by clicking. Then there are some minor stuff I think.

First, this component is not accessible. I couldn't set a rating by keyboard. Then if we are fine with that, the stars make the mouse cursor a pointer, giving a very "clickable" vibe. 

I'd enable setting by clicking and improve accessibility and keep the hover for just highlighting, and not for setting. What do you think?

2pacalypse:
There is also no way to give 0 stars after hovering if that matters

image

src/Nordea/Components/FiveStarRating.elm
sarastorbjork: Clicking is needed to "submit" the rating, but I didn't see a reason to specify here what onClick does (I tried that but then I have to specify the function type, e.g Int -> msg). Instead this can be specified in the view-function (in the same way we do with nordea buttons). Here is how I want to submut the rating in my code https://github.com/SGFinansAS/FinansFront/pull/8539/files . I think this also works then with enter-key (from an accessibility standpoint)?

wenwei-lai (approved):
LGTM :rocket:

src/Nordea/Components/FiveStarRating.elm
sarastorbjork: Hmm it is possible to submit a rating with keyboard, but yes the hover effect doesn't work, I'll work on this !

2pacalypse: Yes, you can enter and submit if you provide `onClick` from outside, but I doubt you can actually "fill" stars with keyboard by tabbing or with arrow keys. I would model this more like

```
type alias Config msg =
    { selectedRating : Int or Maybe Int
    , onClickRating: Int -> msg
    }
```

If you wanna submit when clicked, then you can do that in the message given from `onClickRating`. In fact, you can add delays there for first showing/setting the value, and then stepping to the next stage. The way we provide `onClick` in the `attrs` has come different to me than the rest of this library.

2pacalypse: Now in the current case also, one can click in between the stars despite cursor being non-pointer and trigger the `onClick` in the attrs. I think accessibility-wise there are problems but of course if our entire UX department :sweat_smile: has agreed that for FF this is good enough, then it is fine.

https://github.com/user-attachments/assets/3db6de84-a74a-49d4-a667-ac5776759201

src/Nordea/Components/FiveStarRating.elm
```diff
@@ -0,0 +1,72 @@
...
+        starIcon index =
+            if index <= currentRating then
+                Icons.filledStar iconStarStyle
+
+            else
+                Icons.star iconStarStyle
+
+        starIconButton : Int -> Styled.Html msg
+        starIconButton value =
+            Styled.button
```
2pacalypse, line 51: not sure if `button` is semantically correct here, i.e. maybe radio input is a better one.

2pacalypse (approved):
nice stuff :rocket:

PR #396, opened 2024-07-16
Support different views of DropdownFilter Item

2pacalypse:
New optional methods that can alter the view: `withChevron, withCross`, and

- `withResultsAlwaysShowing`: When focus is lost, results will still show. 
- `withResultsOverlaying`: Whether the result list will be on top of everything -- like in a tooltip way or not. You may wanna set it to False if you want the results to occupy space in your layout. 
- `withScroll`: When set to false, there is no `max-height` limitations, so your list can grow to whatever number of elements without scroll.
- `withItemViewMapper`: In the existing version, one can only show `item.text`. With this method, one can define views based on `item.value` as well.
- `withItemContainerStyles`: This makes it possible to style the item container. Maybe you want larger rows, different borders, change the hover, focus states etc.

An example where most of above things are combined. This is the PH organization search. With this PR, we can remove custom code from PH and use this component while still providing the custom-style via `withItemViewMapper` and `withItemContainerStyles` from PH.

Screenshot 2024-07-22 08:35:14

image

Possible future work: 
- 0 row result treatment 
- isFailure treatment (we have isLoading here that shows a spinner. isFailure makes sense too)
- Work around headers -- especially custom styles for headers.


src/Nordea/Components/DropdownFilter.elm
```diff
@@ -99,9 +57,14 @@ type alias DropdownFilterProperties a msg =
     , hasError : Bool
     , isLoading : Bool
     , hasSearchIcon : Bool
-    , selectedValue : Maybe a
+    , selectedValue : Maybe (Item a)
```
2pacalypse, line 60: This change is breaking, but I have the PRs ready for FF, PH. I can do KP too. It is an easy change.

fredrikloberg (approved):
Nice work!

I tested it with voice over and think accessibility is still lacking a bit. A list of divs doesnt give any hints about being selectable, and is semantically disconnected from the search input. I think this is easily fixed by following the example here https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/combobox_role. The challenge then may be the use of the ID attribute, but if that is required I thing we can wrap the component in a shadow dom and give it a static id. Afaik, IDs must be doesnt need to be unique across shadow dom boundaries https://developer.mozilla.org/en-US/docs/Web/API/Web_components/Using_shadow_DOM#declaratively_with_html

Wdyt?

fredrikloberg (commented):
Nitpick: the clear button isnt perfectly centered vertically. It is centered using absolute positioning, so I guess the parent is lacking a `position: relative`. 

I removed the centering from the screenshot here to illustrate where `top: 0` is:

Screenshot 2024-07-25 at 11 13 18

src/Nordea/Components/DropdownFilter.elm
fredrikloberg: Do we need this change? I guess we could get the `item` from the `a` by searching through the `items` list. That way, we can avoid saving the translation text in the [model in PH](https://github.com/SGFinansAS/PartnerHub/pull/2689/files#diff-464277551852faee39e9bae9bfc824bee95be95fe70b59fd9fa928e83aa4d9c5R25). Not a big deal tho

src/Nordea/Components/DropdownFilter.elm
```diff
@@ -129,97 +155,87 @@ init { onInput, input, onSelect, items, selectedValue } =
...
 
 view : List (Html.Attribute msg) -> DropdownFilter a msg -> Html msg
 view attrs ((DropdownFilter config) as dropdown) =
     let
-        filterValues value =
-            value
-                |> String.toLower
-                |> String.contains (String.toLower config.input)
+        defaultMapper =
+            \item -> Html.text item.text
```
fredrikloberg, line 180:
```suggestion
        defaultMapper item =
            Html.text item.text
```

2pacalypse:
> Nice work!
> 
> I tested it with voice over and think accessibility is still lacking a bit. A list of divs doesnt give any hints about being selectable, and is semantically disconnected from the search input. I think this is easily fixed by following the example here https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/combobox_role. The challenge then may be the use of the ID attribute, but if that is required I think we can wrap the component in a shadow dom and give it a static id. Afaik, IDs doesnt need to be unique across shadow dom boundaries https://developer.mozilla.org/en-US/docs/Web/API/Web_components/Using_shadow_DOM#declaratively_with_html
> 
> Wdyt?

I used the default ubuntu screen reader (Orca), and also tried to look at how google.com did their combobox with autocomplete. It was mostly aligning with their search but they probably don't need "non/selected" stuff to get pronounced.

Added some extra aria stuff now that at least pronounces "non-selected", but to make it pronounce the "selected" I think it requires an `input` and not a `li`. I couldn't get it to pronounce `selected` with `li` (or maybe according to [link ](https://stackoverflow.com/questions/71468332/select2-li-aria-selected-true-is-not-read-as-selected-by-screen-reader) silence means selected?).  For the default vanilla DropdownFilter, sure we can still use an `input` and it then says "selected", but in the more complex cases like the organization search, we then should probably embed/hide an input inside the selected element. I think we already do that in `TexInput` maybe?

https://stackoverflow.com/questions/19236778/how-to-put-div-inside-input-type-text



The accessibility for components deserves more PRs because I think there is whole another story with arrow keys for these components.

Also, `TextInput` (used also in DropdownFilter) is using the given attrs two times. This is leading to repetition of pronunciation for me. This should be improved too in some way.
image

fredrikloberg:
It works better now. But yea we can improve this further in later PRs.

On mac, the screen reader says the following once you focus the text input. So it says there is a combobox inside a combobox, does this make sense? I tried it in elm-explorer
Screenshot 2024-07-25 at 13 50 20

src/Nordea/Components/DropdownFilter.elm
2pacalypse: Hmm, true. I might do that search here and see how it will look.

2pacalypse:
> It works better now. But yea we can improve this further in later PRs.
> 
> On mac, the screen reader says the following once you focus the text input. So it says there is a combobox inside a combobox, does this make sense? 

I think the combobox getting pronounced twice is due to `TextInput` using the same attrs twice.  So that is wrong or a repetition. That should be fixed in some way.

src/Nordea/Components/DropdownFilter.elm
2pacalypse: Alright so keep it backwards compatible I applied this change but in general interface feels a bit awkward because of the mismatch between msg and stored value. They should be either `Maybe a` or `Maybe (Item a)`  together

```
init :
    { onInput : String -> msg
    , input : String
    , onSelect : Maybe (Item a) -> msg
    , selectedValue : Maybe a
    , items : List (ItemGroup a)
    }
```

fredrikloberg: Ah yes I agree. I vote for `Maybe a`

PR #397, opened 2024-07-25
Make chat scroll to bottom by default

fredrikloberg:
Minor fix to get the chat to scroll to bottom by default. So that the user doesnt have to scroll manually when writing a msg.

This is a css only solution using `flex-direction: column-reverse`. See https://stackoverflow.com/a/44051405


2pacalypse (approved):
good renaming of fields also :rocket:

PR #403, opened 2024-08-08
Fix - onEnterPress keycode 

sarastorbjork:
Bug fix where the `onEnterPress` event listened to Spacebar instead of Enter-key.


fredrikloberg (approved):

hhaga (approved):
🙌

PR #408, opened 2024-08-22
Add a unified FileUpload view for single file uploads

rogerab79:
- Adds a unified FileUpload view for single file uploads (with RemoteData for upload state) .
- This also fixes a bug where Decode errors out when added files were not on the accepted list - allowing the drop event to bubble to the top, showing the file contents in the browser.
-  Still not 100% compatible with CP design docs; need to discuss cross-team that all have the same design req.

Breaks API: onFilesSelected handler gives a List that can be empty.

## After file add
image

### DIFF
--- a/src/Nordea/Components/FileUpload.elm
+++ b/src/Nordea/Components/FileUpload.elm
@@ -19,14 +19,14 @@ import Css exposing (absolute, alignItems, backgroundColor, border3, borderRadiu
 import Css.Global as Global
 import File exposing (File)
 import Html.Styled as Html exposing (Attribute, Html)
-import Html.Styled.Attributes exposing (accept, css, multiple, type_, value)
+import Html.Styled.Attributes as Attributes exposing (accept, css, multiple, type_, value)
 import Html.Styled.Events as Events
 import Json.Decode as Decode
 import Maybe.Extra as Maybe
 import Nordea.Components.Spinner as Spinner
 import Nordea.Components.Text as Text
 import Nordea.Components.Tooltip as Tooltip
-import Nordea.Html as Html exposing (showIf, styleIf)
+import Nordea.Html as Html exposing (attrIf, showIf, styleIf)
 import Nordea.Resources.Colors as Colors
 import Nordea.Resources.I18N exposing (Translation)
 import Nordea.Resources.Icons as Icon
@@ -206,36 +206,44 @@ view (FileUpload config) =
                         |> String.join ","
 
         inputFile =
+            let
+                isDisabled =
+                    (config.allowMultiple || List.isEmpty config.files) |> not
+            in
             Html.input
                 [ type_ "file"
                 , onSelectFiles config.accept config.onFilesSelected
                 , multiple config.allowMultiple
                 , accept mimeTypesAsString
                 , value ""
+                , Attributes.disabled True |> attrIf isDisabled
+                , Attributes.attribute "aria-description" (config.translate strings.ariaNoMoreFiles) |> attrIf isDisabled
                 , css [ opacity (num 0), position absolute, height (rem 0), width (rem 0) ]
                 ]
                 []
-                |> showIf (config.allowMultiple || List.isEmpty config.files)
     in
     -- Inline is currently only supported with one file; feel free to enhance :)
     config.onFileRemoval
         |> Maybe.filter (\_ -> not config.allowMultiple && (List.isEmpty config.files |> not) && config.appearance == Large)
         |> Maybe.map
             (\onFileRemoval ->
-                fileList
-                    config.files
-                    onFileRemoval
-                    config.translate
-                    (attrs
-                        ++ [ css
-                                [ borderWidth (px 0) |> Css.important
-                                , cursor default |> Css.important
-                                , Global.children [ Global.typeSelector ":first-child" [ height inherit, width (pct 100) ] ]
+                Html.div []
+                    [ inputFile
+                    , fileList
+                        config.files
+                        onFileRemoval
+                        config.translate
+                        (attrs
+                            ++ [ css
+                                    [ borderWidth (px 0) |> Css.important
+                                    , cursor default |> Css.important
+                                    , Global.children [ Global.typeSelector ":first-child" [ height inherit, width (pct 100) ] ]
 
-                                -- 2024-08-22 Elm css first-child is b0rken! ^
-                                ]
-                           ]
-                    )
+                                    -- 2024-08-22 Elm css first-child is b0rken! ^
+                                    ]
+                               ]
+                        )
+                    ]
             )
         |> Maybe.withDefault
             (Html.div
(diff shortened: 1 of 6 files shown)


src/Nordea/Components/FileUpload.elm
```diff
@@ -242,30 +204,87 @@ view (FileUpload config) =
...
+                    config.files
+                    onFileRemoval
+                    config.translate
+                    (attrs
+                        ++ [ css
+                                [ borderWidth (px 0) |> Css.important
+                                , cursor default |> Css.important
+                                , Global.children [ Global.typeSelector ":first-child" [ height inherit, width (pct 100) ] ]
+
+                                -- 2024-08-22 Elm css first-child is b0rken! ^
```
fredrikloberg, line 235: How is it broken?

rogerab79: It generates a css selector that looks something like this (from memory):
```css
.classhash:first-child {
 ..
}
```
when we want something like
```css
.classhash > :first-child {
  ..
}
```

fredrikloberg: Ah yes true. So I think it would work if you moved the css from the parent to the child then. But yea sometimes it is cleaner to keep it on the parent

rogerab79: Yeah, it would have been the other option; this was kind of an attempt to keep the other code as un-changed as possible avoiding breaking other users :smile:

src/Nordea/Components/FileUpload.elm
```diff
@@ -242,30 +204,87 @@ view (FileUpload config) =
                     mimeTypes
                         |> List.map toMimeTypeString
                         |> String.join ","
+
+        inputFile =
+            Html.input
```
fredrikloberg, line 209: So when you are using this FileUpload component, you are wrapping it in a label of type `InputLabel`. That is fine (and also required) as long as the `FileUpload` component actually contains an input element (textarea, select, and output tags are also valid). But it looks like this component now may not contain any of those, meaning it would be semantically incorrect I think.

We have made attempts of making our Label component support labelling generic content, but then the type of it must be `TextLabel` instead of `InputLabel`.


The components is also initially created with the ideas from [this article](https://www.bekk.christmas/post/2018/10/creating-composite-components) in mind. Instead of having large complex components, we make small composable components.

rogerab79: So the takeout is, if I understand it correctly, that there should be a top component that does the choosing of showing a file listing or the drop zone/file selector then?

fredrikloberg: Yes could be. Just highlighting some potential issues with html semantics and accessibility, I find it very hard to design good components with that in mind. 

The way we have done it previously following the principles from that article, is to keep that logic in the app, where the components are being used. For example: 

```elm
if noFilesSelected then
    Label.init ("Some file document: " ++ remoteText) Label.InputLabel
        |> Label.withRequirednessHint (Just (Label.Optional .no))
        |> Label.view []
            [ FileUpload.init .no OnFilesSelected OnDragEnterFileUpload OnDragLeaveFileUpload
                |> FileUpload.withIsHovering model.customModel.isHoveringFileUpload
                |> FileUpload.withAcceptedFileTypes supportedFileTypes
                |> FileUpload.view
            ]
else 
    Label.init ("Some file document: " ++ remoteText) Label.TextLabel
        |> Label.withRequirednessHint (Just (Label.Optional .no))
        |> Label.view []
            [ SomeFilesUploadedListComponent.init files OnRemove
                |> SomeFilesUploadedListComponent.view []
            ]
```


Or we could make the FileUpload component take in a list of children as sub components, etc. The sub components can be exposed from the same file. The benefit of this approach is that we keep the components as small and simple as possible. And when the designer comes with some fancy new design, we often dont have to alter the components at all, just compose them in a different way in the app.

rogerab79: So a few thoughts about this, having slept on it.
- I need a component that doesn't change size etc. when the file is added. This can obviously be done by cargo-culting the CSS attributes to the other component, but you probably guess by the name-calling how much I dislike that approach.
- I actually don't think it's a "good" thing that elm-components don't need to be changed when a new design arrive, as this means the changes must be applied on each and every project separately. However, a possible design could be having two layers of components; bare unstyled components that only incorporate the logic and the styled components that follow the Nordea styling, whatever it might be at the moment. But realistically, we have this architecture now and new components seldomly arrive so I think we're stuck with this model.

 ... so the alternatives boil down to
1. make a new top component that has the container layout, possibly with appearance in it and then does basically the above `if ... else` to a sub component with some hints. It's a major rewrite/addition of a totally new `SuperFileUpload` with supporting components `FileDrop` and `FileList`. This would duplicate much of the code as FileUpload still would be used for quite a while (ok, FileUpload could internally use `FileDrop`). It would be a *major* hassle however.
2. .or, the current approach, keeping the current component which is basically just minimal additions. If we could continue on this approach, I can also look into bringing more aria hints into the component; e.g. file removal would need a hint, the file input element should be added with the file listing - disabled and possibly with aria hints of why.
3. (and as a last resort, what I really would want to avoid, making this component internally in CP, leaving everybody to re-implement this feature as they will need it in their design)

What specifically would be needed to get option 2 to work? Or which path should I pursuit, in you opinion?

fredrikloberg: > * I actually don't think it's a "good" thing that elm-components don't need to be changed when a new design arrive, as this means the changes must be applied on each and every project separately. However, a possible design could be having two layers of components; bare unstyled components that only incorporate the logic and the styled components that follow the Nordea styling, whatever it might be at the moment. But realistically, we have this architecture now and new components seldomly arrive so I think we're stuck with this model.

Yes I agree there are definitely tradeoffs. We have had quite a few cases where the designer wanted a _slightly_ different design between the apps. For example, a sub component shouldn't be there in one app, one app needs an extra sub component in between, one app needs some extra styling on a specific sub component, etc. So the thought has been that we try to design the components such that it is possible to make app-specific modifications without riddling the components with complex conditionals. While at the same time trying to get the benefit of only changing something one place when there are design changes that should affect every app.

That said, we have had bugs with both approaches. One approach often leads to forgetting to apply a patch to a specific app, while the other approach leads to breaking one app because another app needed a (often app-specific) change to the component.


> What specifically would be needed to get option 2 to work? Or which path should I pursuit, in you opinion?

By going with option 2, we must remember to change the type of the wrapping label when the FileUpload component removes the input tag. Def. possible, and maybe the most pragmatic approach for now? The drawback is that it is easy to forget, and the parent has to know that the FileUpload component internally changes the html based on certain input, and act on it.


My current stance is that writing semantically correct and accessibly html is "easy" when you are writing raw html manually, but hard when you are dealing with reusable composite components. Which is likely because the components are abstracting at the wrong level.

rogerab79: I'll take that as option 2 is possible with some changes; So here are aria + stern usage documentation.
Might have gone a little overboard with aria.. :smile: 

Anything else that is needed for this to be included?

fredrikloberg (approved):
Looks good!

PR #414, opened 2024-09-11
Make chat sender optional

anisha-deverashetty:
In some cases we might not want to show the sender name and only show the message channel. So changed sender to option type.

image


fredrikloberg (approved):

sarastorbjork (commented):
LGTM

PR #422, opened 2024-10-11
Add htmlTitle to Card

morteako:
Adds the possibility to have html as the title for a Card, not just a string

Needed for :
image

image

Backwards comp. (ish), since 
changes should only affect the collapsible card, which is only 1 case of usage, in PH, which will be fixed manually.

For the non-collaps,

```
(old) 
withTitle title_ (Card config) =
    Card { config | title = Just title_ }
and then
in the view
((config.title |> viewMaybe (\title_ -> header [] [ title [] [ Html.text title_ ] ]))

will be the same as

withTitle title_ (Card config) =
    Card { config | htmlTitle = Just () }
and then
((config.htmlTitle |> viewMaybe (\htmlTitle -> htmlTitle))
which is 
((config.htmlTitle |> viewMaybe (\htmlTitle -> header [] [ title [] [ Html.text htmlTitle ]]))

which is the same

### CODE REVIEW

File: src/Nordea/Components/Card.elm [hunk 1]
```diff
@@ -108,6 +111,7 @@ view attrs children (Card config) =
     else
         Html.div baseStyle
             ((config.title |> viewMaybe (\title_ -> header [] [ title [] [ Html.text title_ ] ]))
+                :: (config.htmlTitle |> viewMaybe (\htmlTitle -> htmlTitle))
```
morteako, line 114: for the String title, it currently adds
marginBottom 1.5 rem.

Should we do this for the htmlTitle also?
F.ex. by wrapping the htmlTitle here in `div [css [marginBottom (rem 1.5)] [htmlTitle]` ?
@fredrikloberg

src/Nordea/Components/Card.elm
```diff
@@ -98,6 +101,7 @@ view attrs children (Card config) =
             baseStyle
             (headerCollapsible (Maybe.values [ config.onClick |> Maybe.map onClick ] ++ attrs)
                 [ config.title |> viewMaybe (\title_ -> Text.bodyTextHeavy |> Text.view [ css [] ] [ Html.text title_ ])
+                , config.htmlTitle |> viewMaybe (\htmlTitle -> Html.div [ css (if config.isOpen then [ marginBottom (rem 1.5) ] else []) ] [ htmlTitle ])
```
morteako, line 104: There is an issue with margin/gap when the card is opened.

image

This is fixed by this conditional marginBottom.
But do we want this here? or should this be controlled by the user? 🤔 
@fredrikloberg maybe knows

explorer/review/src/ReviewConfig.elm
```diff
@@ -14,15 +14,15 @@ For using elm-review and adding new rules see -> <https://github.com/jfmengels/n
 -}
 
 import NoBooleanCase
+import NoConfusingPrefixOperator
```
morteako, line 17: got formatted by elm-format. Should just keep it

src/Nordea/Components/Card.elm
```diff
@@ -53,6 +54,7 @@ import Nordea.Resources.Icons as Icons
 
 type alias CardProperties msg =
     { title : Maybe String
+    , htmlTitle : Maybe (Html msg)
```
fredrikloberg, line 57: What about reducing the setTitle function to set the string title as a htmlTitle? Then we avoid state

src/Nordea/Components/Card.elm
fredrikloberg: Hmm, what if you move it to the content container as a marginTop instead, can the conditional be removed then?

src/Nordea/Components/Card.elm
morteako: i was thinking about it. I guess it depends on how the gap-stuff is dealt with, to avoid UI-breakage for all other users.

src/Nordea/Components/Card.elm
fredrikloberg: I remember wanting the title to be of type html when creating this component, but that got shut down. So the result is now that we have quite a lot of custom logic for this card spread around. So I think the margin stuff has to be tested

fredrikloberg (approved):

src/Nordea/Components/Card.elm
morteako: Yes. I actually see now that the current title-stuff is "broken", in that the margin is weird there also.
image

The issue is the actually the padding

image

PR #427, opened 2024-11-12
Update Badge component

anisha-deverashetty:
Updated Badge component to have more variations. Updated the design for Generic type

**Badge without icon**
image
**Badge with max char count**
image
**Generic**
image

### DIFF
--- a/explorer/src/Stories/Badge.elm
+++ b/explorer/src/Stories/Badge.elm
@@ -37,14 +37,6 @@ stories =
                     ]
           , {}
           )
-        , ( "Generic"
-          , \_ ->
-                Html.div [ css [ displayFlex ] ]
-                    [ Badge.Generic
-                        |> Badge.view [] [ Icons.pdf [ css [ width (rem 5) |> Css.important ] ] ]
-                    ]
-          , {}
-          )
         , ( "Number no icon"
           , \_ ->
                 Html.div [ css [ displayFlex ] ]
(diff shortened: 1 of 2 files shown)


explorer/src/Stories/Badge.elm
```diff
@@ -45,4 +45,28 @@ stories =
...
+                Html.div [ css [ displayFlex ] ]
+                    [ Badge.Dot |> Badge.view [] [ Icons.calendar [ css [ width (rem 2) |> Css.important ] ] ] ]
+          , {}
+          )
+        , ( "Dot no icon"
+          , \_ ->
+                Html.div [ css [ displayFlex ] ]
+                    [ Badge.Dot |> Badge.view [] [] ]
+          , {}
+          )
```
fredrikloberg, line 71: Do we need this dot variant? Best if we could keep it simple and only have one generic variant

anisha-deverashetty: We have the design here. The dot is the one we need to use when we don't have the number. I couldn't find the generic variant in design lib and Line too didn't know about this one. 
image

anisha-deverashetty: If we are not using this generic one then it should be replaced to dot. What you think?

fredrikloberg: Yes I agree, challenge Line about replacing the old generic one rather than adding a new

PR #429, opened 2024-11-20
Workaround for Chrome bug in details element

kjesvale:
This is a workaround for a bug in Chrome that messes up the Grid layout when using the HTML details element together with the `display: contents` rule. The bug is currently apparent on the Supplier Invoice and Change of Party pages in FinansFront.

Fixed as described on the Chrome bug tracking site: https://issues.chromium.org/issues/40069963

Before:

image

After:

image


fredrikloberg (approved):

trysnetron (approved):
LHTM :t-rex:

PR #431, opened 2024-11-25
Make emphasised text optional in card

anisha-deverashetty:
Make emphasised text optional in collapsible card.


hhaga (approved):
💯  Great work!

PR #432, opened 2024-11-25
Initial Utils

morteako:
* Adds Utils-folder
* Adds Utils/List.elm with isNotEmpty-function and use it where appropriate

Not 100% sure about the Path/module name.
Currently it is placed inside Nordea,
so the import will be `import Nordea.Utils.List as List`
Could also be put outside and be `import Utils.List as List` but that can crash with local util dirs i guess?
Can be nice to easily see that is from component library with the Nordea-prefix


fonnes (approved):

trysnetron (approved):
LGTM :rocket:

PR #433, opened 2024-11-27
Fix not empty conditions

anisha-deverashetty:
The conditions were updated wrongly in this PR https://github.com/SGFinansAS/elm-components/pull/432
**src/Nordea/Components/Chat.elm**
image
**src/Nordea/Components/FileUpload.elm**
image

### DIFF
--- a/src/Nordea/Components/FileUpload.elm
+++ b/src/Nordea/Components/FileUpload.elm
@@ -209,7 +209,7 @@ view (FileUpload config) =
         inputFile =
             let
                 isDisabled =
-                    config.allowMultiple || List.isNotEmpty config.files
+                    not config.allowMultiple || List.isNotEmpty config.files
             in
             Html.input
                 [ type_ "file"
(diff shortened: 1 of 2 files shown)


src/Nordea/Components/FileUpload.elm
```diff
@@ -209,7 +209,7 @@ view (FileUpload config) =
         inputFile =
             let
                 isDisabled =
-                    config.allowMultiple || List.isNotEmpty config.files
+                    not config.allowMultiple || List.isNotEmpty config.files
```
fredrikloberg, line 212: Given how it was before, I think we must change it to && for it to be the same?

```suggestion
                    not config.allowMultiple && List.isNotEmpty config.files
```

anisha-deverashetty: Yes, updated now 🫣

fredrikloberg (approved):

PR #435, opened 2024-12-10
Avoid jerking the UI when showing error message

rogerab79:
Currently Label doesn't reserve space for the error message resulting in a noticable UI jerk when the error message takes up new space. Fix this by reserving space for the error message using visibility: hidden, also only do this if withError is used to avoid noticeable changes for other users.

Also 
- Enhance stories so that the problem can be reproduced in-tree.
- Add a cssIf helper, couldn't find anything matching this; holler if you know an existing implementation!

## UI Jerk in action
https://github.com/user-attachments/assets/bd27f029-4018-415c-b3c2-3e6817565fb4

### DIFF
--- a/explorer/src/Config.elm
+++ b/explorer/src/Config.elm
@@ -49,6 +49,7 @@ type alias Config =
     , showCoachMarkStep : Maybe Int
     , activeRadioButton : String
     , textInputContent : String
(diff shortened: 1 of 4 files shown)


fredrikljung93 (approved):
LGTM :rocket:

PR #444, opened 2025-02-10
Add Icons: Sign method logos

morteako:
image

also rename from
Icons.norway to Icons.flagNorway, since it is a flag :))


fonnes (approved):
Llgstm

trysnetron (approved):
LGTM🚀

PR #451, opened 2025-02-19
Introduce helper to opt out of reserving space for errors on labels

kjesvale:
## Edit

Introduce a `Label.withNoReservedErrorSpace` for opting out of the current behaviour that reserves space below the Label component if using `Label.withErrorMessage Nothing`

---

When using `Label.withErrorMessage`, the Label component reserves space for the error message, even if the provided error message is `Nothing`. This makes it difficult to control spacing between components in a form.

I'm not sure if this behaviour is a feature or a bug. This PR proposes a bugfix.

### ✅ No error message:

Works as expected.

```elm
Label.init "No error message" Label.InputLabel
```

image

### ✅ With error message = Just "Message"

Works as expected.

```elm
Label.init "With error message" Label.InputLabel
    |> Label.withErrorMessage (Just "Message")
```

image

### ❌ With error message = Nothing

Notice the extra space between the second and third field.

```elm
Label.init "With empty error message" Label.InputLabel
    |> Label.withErrorMessage Nothing
```

image

### ✅ With error message = Nothing

No extra space between the second and third component when using `Label.withErrorMessage Nothing`.

image

### DIFF
--- a/src/Nordea/Components/Label.elm
+++ b/src/Nordea/Components/Label.elm
@@ -49,6 +49,7 @@ type alias InputProperties =
     , labelType : LabelType
     , requirednessHint : Maybe RequirednessHint
     , errorMessage : Maybe String
+    , withError : Bool
     , reserveSpaceForError : Bool
     , hintText : Maybe String
     , charCounter : Maybe CharCounter
@@ -72,7 +73,8 @@ init labelText labelType =
         , labelType = labelType
         , requirednessHint = Nothing
         , errorMessage = Nothing
-        , reserveSpaceForError = False
+        , withError = False
+        , reserveSpaceForError = True
         , hintText = Nothing
         , charCounter = Nothing
         , size = Standard
@@ -140,7 +142,7 @@ view attrs children (Label config) =
                             [ String.fromInt counter.current ++ "/" ++ String.fromInt counter.max |> Html.text ]
 
                 showErrorRow =
-                    config.errorMessage /= Nothing || config.reserveSpaceForError
+                    config.withError && config.reserveSpaceForError
             in
             Html.row
                 (css
@@ -252,7 +254,7 @@ withRequirednessHint requirednessHint (Label config) =
 
 withErrorMessage : Maybe String -> Label -> Label
 withErrorMessage errorMessage (Label config) =
-    Label { config | errorMessage = errorMessage, reserveSpaceForError = True }
+    Label { config | errorMessage = errorMessage, withError = True }
 
 
 withNoReservedErrorSpace : Label -> Label
--- a/src/Nordea/Components/Label.elm
+++ b/src/Nordea/Components/Label.elm
@@ -255,8 +255,8 @@ withErrorMessage errorMessage (Label config) =
     Label { config | errorMessage = errorMessage, reserveSpaceForError = True }
 
 
-withNoReservedErrorSpace : Bool -> Label -> Label
-withNoReservedErrorSpace reserveSpace (Label config) =
+withNoReservedErrorSpace : Label -> Label
+withNoReservedErrorSpace (Label config) =
     Label { config | reserveSpaceForError = False }
 
 
--- a/src/Nordea/Components/Label.elm
+++ b/src/Nordea/Components/Label.elm
@@ -48,7 +48,6 @@ type alias InputProperties =
     , labelType : LabelType
     , requirednessHint : Maybe RequirednessHint
     , errorMessage : Maybe String
-    , withError : Bool
     , hintText : Maybe String
     , charCounter : Maybe CharCounter
     , size : Size
@@ -71,7 +70,6 @@ init labelText labelType =
         , labelType = labelType
         , requirednessHint = Nothing
         , errorMessage = Nothing
-        , withError = False
         , hintText = Nothing
         , charCounter = Nothing
         , size = Standard
@@ -119,17 +117,14 @@ view attrs children (Label config) =
                     initText config.size
                         |> Text.view [ css [ color Colors.darkGray ] ] [ Html.text text ]
 
-                viewError maybeError =
+                viewError error =
                     initText config.size
                         |> Text.view
                             [ class "input-focus-color"
-                            , css
-                                [ displayFlex
-                                , visibility hidden |> Css.cssIf (Maybe.isNothing maybeError)
-                                ]
+                            , css [ displayFlex ]
                             ]
                             [ Icons.error [ css [ marginRight (rem 0.5), flex none ] ]
-                            , maybeError |> Maybe.withDefault "No errors" |> Html.text
+                            , Html.text error
                             ]
 
                 viewCharCounter counter =
@@ -146,12 +141,12 @@ view attrs children (Label config) =
                     :: attrs_
                 )
                 [ Html.column []
-                    [ viewError config.errorMessage |> Html.showIf config.withError
+                    [ Html.viewMaybe viewError config.errorMessage
                     , config.hintText |> Html.viewMaybe viewHintText
                     ]
                 , config.charCounter |> Html.viewMaybe viewCharCounter
                 ]
-                |> showIf (config.withError || config.hintText /= Nothing || config.charCounter /= Nothing)
+                |> showIf (config.errorMessage /= Nothing || config.hintText /= Nothing || config.charCounter /= Nothing)
     in
     case config.labelType of
         InputLabel ->
@@ -248,7 +243,7 @@ withRequirednessHint requirednessHint (Label config) =
 
 withErrorMessage : Maybe String -> Label -> Label
 withErrorMessage errorMessage (Label config) =
-    Label { config | errorMessage = errorMessage, withError = True }
+    Label { config | errorMessage = errorMessage }
 
 
 withHintText : Maybe String -> Label -> Label
--- a/src/Nordea/Components/Label.elm
+++ b/src/Nordea/Components/Label.elm
@@ -7,6 +7,7 @@ module Nordea.Components.Label exposing
     , withCharCounter
     , withErrorMessage
     , withHintText
+    , withNoReservedErrorSpace
     , withRequirednessHint
     , withSmallSize
     )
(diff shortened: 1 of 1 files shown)


fredrikljung93:
An advantage with reserving that space could be avoiding the other elements from "jumping around the screen" as the error message appears and disappears :thinking:

fredrikloberg:
> An advantage with reserving that space could be avoiding the other elements from "jumping around the screen" as the error message appears and disappears 🤔

Yes so I think this was the reasoning for reserving the space

fredrikloberg:
Another point is when you have a flowing flex layout, you may end up with unalignment like this:
Screenshot 2025-02-19 at 13 39 38

vs when you reserve space:
Screenshot 2025-02-19 at 13 39 58

kjesvale:
> Another point is when you have a flowing flex layout, you may end up with unalignment like this

Looks like that one can be solved by using a CSS Grid instead? I'm aware of the behaviour that elements below the error-inducing component will be pushed down. But if you align the components using CSS, I don't see this as an issue

kjesvale:
> An advantage with reserving that space could be avoiding the other elements from "jumping around the screen" as the error message appears and disappears 🤔

When you get an error message, it is usually in response to some user action, and the user expects something to change anyway. I don't think a small jump within the form is a bad thing in this case.

fredrikloberg:
> Looks like that one can be solved by using a CSS Grid instead? I'm aware of the behaviour that elements below the error-inducing component will be pushed down. But if you align the components using CSS, I don't see this as an issue
> 
> I think the last screenshot also looks a little bit weird since the spacing varies between the fields with and without an error message

True, can most probably be solved in other ways. But what do you think about the jankiness? You we avoid that?

fredrikloberg:
> When you get an error message, it is usually in response to some user action, and the user expects something to change anyway. I don't think a small jump within the form is a bad thing in this case.

Im leaning the other way. I think the jankiness can be very disturbing. The error message will appear the first time on blur, so the user could click out to another field and then suddenly the whole page jumps down. In PH for example, one field affects the validation status in other fields too, so it is not just jumping the height of one error message, but it could be the height of two or three combined.

kjesvale:
> > Looks like that one can be solved by using a CSS Grid instead? I'm aware of the behaviour that elements below the error-inducing component will be pushed down. But if you align the components using CSS, I don't see this as an issue
> > I think the last screenshot also looks a little bit weird since the spacing varies between the fields with and without an error message
> 
> True, can most probably be solved in other ways. But what do you think about the jankiness? Can we avoid that?

As I responded to Ljung, I don't think the content shifting is janky when in response to the user's action. I agree content shifting should be avoided in a lot of cases, for instance when loading the page initially, but not necessarily when the user is interacting with the form

rogerab79:
> > > Looks like that one can be solved by using a CSS Grid instead? I'm aware of the behaviour that elements below the error-inducing component will be pushed down. But if you align the components using CSS, I don't see this as an issue
> > > I think the last screenshot also looks a little bit weird since the spacing varies between the fields with and without an error message
> > 
> > 
> > True, can most probably be solved in other ways. But what do you think about the jankiness? Can we avoid that?
> 
> As I responded to Ljung, I don't think the content shifting is janky when in response to the user's action. I agree content shifting should be avoided in a lot of cases, for instance when loading the page initially, but not necessarily when the user is interacting with the form

I'm the "guilty" one adding this space behavior, and yes, the UI jerking was really annoying and would have required an external component reserving the space otherwise. If you still want the space-less behavior in some place, you could just add a function that is identity whenever there is no error instead of doing withError. This may be "ok"-ish if you end up refreshing the whole page and moving stuff around between the errors, but if you are yanking the UI just a tiny bit, f.ex when the user types into the field, you will most likely annoy the user..

Disclaimer: I really, **REALLY** hate UI:s that jerk around as I interact with them. You should too! :laughing:

kjesvale:
> > > > Looks like that one can be solved by using a CSS Grid instead? I'm aware of the behaviour that elements below the error-inducing component will be pushed down. But if you align the components using CSS, I don't see this as an issue
> > > > I think the last screenshot also looks a little bit weird since the spacing varies between the fields with and without an error message
> > > 
> > > 
> > > True, can most probably be solved in other ways. But what do you think about the jankiness? Can we avoid that?
> > 
> > 
> > As I responded to Ljung, I don't think the content shifting is janky when in response to the user's action. I agree content shifting should be avoided in a lot of cases, for instance when loading the page initially, but not necessarily when the user is interacting with the form
> 
> I'm the "guilty" one adding this space behavior, and yes, the UI jerking was really annoying and would have required an external component reserving the space otherwise. If you still want the space-less behavior in some place, you could just add a function that is identity whenever there is no error instead of doing withError. This may be "ok"-ish if you end up refreshing the whole page and moving stuff around between the errors, but if you are yanking the UI just a tiny bit, f.ex when the user types into the field, you will most likely annoy the user..
> 
> Disclaimer: I really, **REALLY** hate UI:s that jerk around as I interact with them. You should too! 😆

Aaah I can see the changes were introduced in December (https://github.com/SGFinansAS/elm-components/pull/435)

Wasn't sure if this was intentional, but I guess this is a matter of opinion then 😄 Or the type of application. I'm coming from FinansFront's point of view, where the interface can have a lot of stuff going on. In such cases, I like to prioritize grouping fields together visually, over the layout shifting that happens if an error occur. For instance, on the supplier invoice page, it's tidy to have "Brutto", "Netto" and "Moms" grouped close together:

image

But I understand that is not always the preferred behaviour.

Would it be possible to make this option more explicit? For instance with a `Label.withShowError` or something that reserves the error space if True?

rogerab79:
> > I'm the "guilty" one adding this space behavior, and yes, the UI jerking was really annoying and would have required an external component reserving the space otherwise. If you still want the space-less behavior in some place, you could just add a function that is identity whenever there is no error instead of doing withError. This may be "ok"-ish if you end up refreshing the whole page and moving stuff around between the errors, but if you are yanking the UI just a tiny bit, f.ex when the user types into the field, you will most likely annoy the user..
> > Disclaimer: I really, **REALLY** hate UI:s that jerk around as I interact with them. You should too! 😆
> 
> Aaah I can see the changes were introduced in December (#435)
> 
> Wasn't sure if this was intentional, but I guess this is a matter of opinion then 😄 Or the type of application. I'm coming from FinansFront's point of view, where the interface can have a lot of stuff going on. In such cases, I like to prioritize grouping fields together visually, over the layout shifting that happens if an error occur. For instance, on the supplier invoice page, it's tidy to have "Brutto", "Netto" and "Moms" grouped close together:
> image
> 
> But I understand that is not always the preferred behavior.
> 
> Would it be possible to make this option more explicit? For instance with a `Label.withShowError` or something that reserves the error space if True?

I see this might be a valid case for allowing jerking of the UI. Even better would be to inline the error, but I guess the FF horizontal real estate is too limited for that.. 
Would a simple fixup helper there work e.g. (pseudo elm, not tested and too lazy to lookup the Label api) 
```elm
helper maybeError = Maybe.map (Just >> Label.withError) maybeError |> Maybe.withDefault identity
-- usage:
Label.init |> helper myError |> Label.view
```

What I'm saying/pondering is; which one will be used more?  Changing the API breaks stuff, and I get the feeling that no-jerk is mostly the preferred behavior. Adding a `withNoReserveError`  for this particular use case might also be a way forward so you can explicitly mark the cases where you know UI jerk is acceptable and the preferred method?

kjesvale:
> I see this might be a valid case for allowing jerking of the UI. Even better would be to inline the error, but I guess the FF horizontal real estate is too limited for that.. Would a simple fixup helper there work e.g. (pseudo elm, not tested and too lazy to lookup the Label api)
> 
> ```elm
> helper maybeError = Maybe.map (Just >> Label.withError) maybeError |> Maybe.withDefault identity
> -- usage:
> Label.init |> helper myError |> Label.view
> ```
> 
> What I'm saying/pondering is; which one will be used more? Changing the API breaks stuff, and I get the feeling that no-jerk is mostly the preferred behavior. Adding a `withNoReserveError` for this particular use case might also be a way forward so you can explicitly mark the cases where you know UI jerk is acceptable and the preferred method?

I checked the guidelines on [Gov.uk](https://govuk-elements.herokuapp.com/form-elements/example-form-elements/), and they put the error message between the label and field, with no reserved space. I also checked the guidelines for the [Norwegian Welfare Administration](https://aksel.nav.no/monster-maler/soknadsdialog/monster-for-skjemavalidering) and they put the error below the field, but without reserving any space.

These are both organizations with a strong emphasis on UX and accessibility. So I don't think it's obvious that reserving the space is the preferred behaviour. Also, wasn't this extra space added to our component library just recently, in December?

My original issue with the extra space is that it becomes difficult to control the spacing between components if you reserve space for errors. This also becomes an even bigger issue if the error message is broken between multiple lines.

At the very least, I think it should be easy to opt-out of this behaviour. I could live with something like a `Label.withNoReserveError` helper as you suggested 😄

rogerab79:
> > What I'm saying/pondering is; which one will be used more? Changing the API breaks stuff, and I get the feeling that no-jerk is mostly the preferred behavior. Adding a `withNoReserveError` for this particular use case might also be a way forward so you can explicitly mark the cases where you know UI jerk is acceptable and the preferred method?
> 
> I checked the guidelines on [Gov.uk](https://govuk-elements.herokuapp.com/form-elements/example-form-elements/), and they put the error message between the label and field, with no reserved space. I also checked the guidelines for the [Norwegian Welfare Administration](https://aksel.nav.no/monster-maler/soknadsdialog/monster-for-skjemavalidering) and they put the error below the field, but without reserving any space.

Oh man that form at aksel.nav.no was IMO terrible from an UX pow - the whole form just blowing up if you press next and you really have to re-navigate to what you were doing at the time only to save some vertical space (that basically would have been blank space). I get it, there are many opinions and even the best studies might disagree on the result of this. And I happen to have a strong opinion of having the layout as static as possible - influenced by the UI/UX courses I've attended. Also, govs are known to have made mistakes at times :laughing: 


> 
> These are both organizations with a strong emphasis on UX and accessibility. So I don't think it's obvious that reserving the space is the preferred behaviour. Also, wasn't this extra space added to our component library just recently, in December?
> 
> My original issue with the extra space is that it becomes difficult to control the spacing between components if you reserve space for errors. This also becomes an even bigger issue if the error message is broken between multiple lines.
> 
> At the very least, I think it should be easy to opt-out of this behaviour. I could live with something like a `Label.withNoReserveError` helper as you suggested 😄

Then it sounds like a plan :rocket: - allowing both worlds to co-exist is probably the best way forward.

rogerab79 (approved):
LGTM!

PR #458, opened 2025-04-01
Group label bottom margin fix

heiditark:
Fixed a small bottom margin difference between InputLabel and GroupLabel.

Before:
image

After :fire::
image


fredrikloberg (approved):
🙌

PR #461, opened 2025-04-03
Add groups to multiselect

morteako:
image

Adds groups to multiselectdropdown. See design on left.

I have not changed the actual design and spacing of the select items, so it is not 100% as the design.

I have kept the current withOptions-functions, so it should be non-breaking


rogerab79 (approved):
LGTM!

PR #462, opened 2025-04-08
Update "OnOutsideClickListener" to support strict CSP header

fredrikloberg:
Update the OnOutsideClickListener to support a strict CSP header. Instead of injecting inline js in an iframe, we must properly import the needed js snippet. The new behaviour is opt-in for now because it is easy to forget importing the js.

Also fixed the fonts in Elm UI explorer


2pacalypse (approved):
:1st_place_medal:

anisha-deverashetty (approved):
Nice :+1:

PR #465, opened 2025-05-07
MultiCompanySelector - add selected before input

morteako:
Add the showing of the selected items, for the multiselect with filter. 
(Not implemented for the non search-filter version. that can be added later)

To fix input focus when clicking the non-input-row, i need to use Browser.Dom.focus from elm/browser. 

### Break works fine wiht multiple 

image

### DIFF
--- a/src/Nordea/Components/MultiSelectDropdown.elm
+++ b/src/Nordea/Components/MultiSelectDropdown.elm
@@ -14,7 +14,7 @@ module Nordea.Components.MultiSelectDropdown exposing
     , withSelected
     )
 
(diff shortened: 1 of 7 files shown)


src/Nordea/Components/MultiSelectDropdown.elm
```diff
@@ -327,6 +387,16 @@ viewSelectItems dropdown index orgOptionGroup =
             )
 
 
+focusInput : InputProperties msg -> Cmd msg
+focusInput inputProperties =
+    Task.attempt (\_ -> inputProperties.noOp) (Browser.focus (makeId inputProperties))
```
fredrikloberg, line 392: Having a noop feels a bit cumbersome. I think it is fine to fire onInput with the current input for example. I mean, you did get some input, kinda like how a space + backspace input would work like a noop.

fredrikloberg (approved):
Looks good

src/Nordea/Components/MultiSelectDropdown.elm
morteako: so i guess it will be
```
onClick (inputProperties.onInput inputProperties.input (
Task.attempt (\_ -> inputProperties.onInput inputProperties.input Cmd.none) (Browser.focus (makeId inputProperties))))
```

😅 @fredrikloberg

PR #469, opened 2025-05-13
MultiSelect - fix search bar icon placement

morteako:
fix search bar icon placement, and also use less hacky non-absolute placement and such

Old:
image

New:
image


rogerab79 (approved):
LGTM!

PR #475, opened 2025-05-27
Fix InstructionWarning svg - remove fixed heigh and width

morteako:
Current makes it impossible(?) to change size.
This makes it responsive :))


fonnes (approved):

trysnetron (approved):
:rocket:

PR #479, opened 2025-06-13
Improve accessibility of text label

fredrikloberg:
Text labels can be semantically described with the definition list element: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/dl


fonnes (approved):
LAHTM

PR #482, opened 2025-06-18
Remove 'withError' on label

fredrikloberg:
The bool currently makes it hard to make a consistent layout when mixing form controls with and without errors.


2pacalypse (approved):

hhaga (approved):
💯

PR #493, opened 2025-08-12
Improve chat history accessibility

fredrikloberg:
Improve chat history accessibility by 
* using ol/li tags for list semantics
* Each message is wrapped in an article tag
* The metadata (who sent, timestamp, etc.) about each message is put in header/footer tags inside the article
* The body of the chat message is put in a p-tag

### DIFF
--- a/src/Nordea/Components/Chat.elm
+++ b/src/Nordea/Components/Chat.elm
@@ -253,7 +253,7 @@ chatHistoryView attrs { sentFrom, sentAt, sender, message, isIncomingMessage, re
                         ]
                         [ Html.text message ]
                 , readReceipt
-                    |> Maybe.map (\r -> Html.footer [] [ messageLabel [ css [ marginTop (rem 0.625) ] ] r ])
+                    |> Maybe.map (\r -> Html.footer [] [ messageLabel [] r ])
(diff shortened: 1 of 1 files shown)


src/Nordea/Components/Chat.elm
```diff
@@ -194,49 +211,52 @@ chatHistoryView attrs { sentFrom, sentAt, sender, message, isIncomingMessage, re
...
-                                    , backgroundColor Colors.lightBlue
-                                    , textOverflow ellipsis
-                                    , overflow hidden
-                                    , whiteSpace noWrap
-                                    ]
-                                ]
-                                [ Html.text messageChipText_ ]
-                    )
-            ]
-        , Html.column [ class "chat-message", css [ padding (rem 0.625), gap (rem 0.625), messageStyles ] ]
```
fredrikloberg, line 226: @anisha-deverashetty , do you know what this class is used for, and if it is ok to remove?

fredrikloberg: I see, used by FF

2pacalypse (approved):

anisha-deverashetty (approved):

PR #498, opened 2025-08-26
Modal with html title

heiditark:
Implemented `withHtmlTitle` to Modal component to be able to have custom titles on the modal. It is not possible to have modal completely without header but title text can be Nothing.

Screenshots:
image

image


fonnes (approved):
Lgtm!

PR #533, opened 2026-02-26
Make Toggle more accessible

rogerab79:
- Rework to the config .with* pattern.
- Allow special quirks for difficult UI patterns.

# Story:
image

# Firefox a11y view
image


fredrikloberg (approved):
Looks good!

PR #550, opened 2026-07-21
Add withHtmlTag to Card

rogerab79:
Allows using e.g. `li` to make shallower hierarchies.
image

### DIFF
--- a/src/Nordea/Components/Card.elm
+++ b/src/Nordea/Components/Card.elm
@@ -6,6 +6,7 @@ module Nordea.Components.Card exposing
     , isCollapsible
     , title
     , view
+    , withHtmlTag
     , withHtmlTitle
     , withShadow
     , withTitle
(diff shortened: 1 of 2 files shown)


src/Nordea/Components/Card.elm
```diff
@@ -61,6 +62,7 @@ type alias CardProperties msg =
     , isCollapsible : Bool
     , isOpen : Bool
     , onClick : Maybe msg
+    , htmlTag : Maybe (List (Attribute msg) -> List (Html msg) -> Html msg)
```
morteako, line 65: i understand the reason for the name, since it basically changes the tag, 
but it feels more like a constructor or functino.

does htmlConstructor make sense? maybe not

morteako, line 65: also possible just have Maybe String, and use the html.node (tag |> withDefault "div")

morteako (approved):

src/Nordea/Components/Card.elm
rogerab79: We already use this pattern, so not sure it is wise to diverge from it (`git grep withHtmlTag`). The config naming also follows..

rogerab79: Currently I don't see a use case for this, so keeping things agile... :laughing:

